PR #1, opened 2024-06-13
chore: initial project structure [part 1/5]

andreabadesso:
### Motivation

We should create the initial project structure, including all dev dependencies

### Acceptance Criteria
- We should have a `flake.nix` defining all system dependencies we need and it should be loaded by `nix-direnv` automatically (through the .envrc)
- The project should be a `yarn workspaces` project, with a single package, the `hathor-rpc-handler`
- The `hathor-rpc-handler` package should include `eslint`, `tseslint` and `typescript` and use fixed versions
- We should use the recommended lint configs from `eslint` and `tseslint`

### Checklist


.github/PULL_REQUEST_TEMPLATE/release_pr_template.md
```diff
@@ -0,0 +1,12 @@
...
+
+Link here all the PRs that are included in this release
+
+# Release-candidates
+
+Link here the release-candidates that were deployed as part of this release
+
+# Checklist
+
+- [ ] I've read and followed the release process described in https://github.com/HathorNetwork/ops-tools/blob/master/docs/release-guides/wallet-service.md#stable-release
```
pedroferreira1, line 11: Link is from wallet service release process

package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "hathor-rpc-lib",
+  "private": true,
```
pedroferreira1, line 3: Why is it private?

flake.nix
```diff
@@ -0,0 +1,42 @@
+{
+  description = "virtual environments";
+
+  inputs = {
+    devshell.url = "github:numtide/devshell";
+    flake-utils.url = "github:numtide/flake-utils";
+    unstableNixPkgs.url = "nixpkgs/nixos-unstable";
```
pedroferreira1, line 7: What's this? Why is it unstable?

packages/hathor-rpc-handler/eslint.config.js
```diff
@@ -0,0 +1,14 @@
...
+
+const eslint = require('@eslint/js');
+const tseslint = require('typescript-eslint');
+
+module.exports = tseslint.config(
+  eslint.configs.recommended,
+  ...tseslint.configs.recommended,
+  {
+    ignores: [
+      'eslint.config.js',
```
pedroferreira1, line 11: Why do you need to ignore this?

package.json
andreabadesso: This is the root package, which will never be published, so it's private

The actual packages that will be build are inside `packages/`

.gitattributes
```diff
@@ -0,0 +1,4 @@
+/.yarn/**            linguist-vendored
```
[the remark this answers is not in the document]
andreabadesso: Yarn created these automatically, they help git, github or whatever git client that is reading them understand what these files are (like not showing a diff when the file is binary

Kind of pointless though as we are adding `.yarn` to `.gitignore`, I'll delete this file

.github/PULL_REQUEST_TEMPLATE/release_pr_template.md
andreabadesso: Removed in 50213095c637848b9edcd20ec49af774453aad4d

flake.nix
andreabadesso: I was using yarn from the unstable nix channel as the latest version didn't have yarn v2 available

It's now available, so we don't need to fetch it from the unstable channel

updated in f5fa3d18ad6aa71a3236dee14ac6aeb8a4a2f6fd

packages/hathor-rpc-handler/eslint.config.js
andreabadesso: Fixed in 61983cabfb8c4b125baca60a8827dcd54ad4e41a

packages/hathor-rpc-handler/src/index.ts
```diff
@@ -0,0 +1,2 @@
+export default () => {
```
tuliomir, line 1: question: Should we have the Hathor header here?

pedroferreira1 (approved):

packages/hathor-rpc-handler/src/index.ts
andreabadesso: Done! Thanks

.gitignore
```diff
@@ -0,0 +1,15 @@
+.yarn/*
+!.yarn/patches
+!.yarn/plugins
+!.yarn/releases
+!.yarn/sdks
+!.yarn/versions
```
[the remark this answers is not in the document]
andreabadesso: Yarn added those automatically, I don't like them at all..

[Their justification](https://yarnpkg.com/features/caching#zero-installs) is that they want everything to be contained in the project, so all devs share the exactly same environment

I'm using nix for this exact purpose and I think it's way superior... 

Updated gitignore to remove those

Thanks

tuliomir (approved):

PR #16, opened 2024-08-21
feat: checking if the requested blueprint exists in the fullnode

andreabadesso:
### Motivation

We should validate the blueprint id before yielding prompts to the clients

### Acceptance Criteria

- We should request the fullnode for the blueprint information and fail if unable to fetch it.

### Checklist

### DIFF
--- a/packages/hathor-rpc-handler/src/errors/index.ts
+++ b/packages/hathor-rpc-handler/src/errors/index.ts
@@ -18,3 +18,7 @@ export class DifferentNetworkError extends Error {};
 export class NoUtxosAvailableError extends Error {};
 
 export class SignMessageError extends Error {};
+
+export class ParamsValidationError extends Error {};
+
+export class BluePrintNotFoundError extends ParamsValidationError {};
--- a/packages/hathor-rpc-handler/src/rpcMethods/sendNanoContractTx.ts
+++ b/packages/hathor-rpc-handler/src/rpcMethods/sendNanoContractTx.ts
@@ -5,7 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import { HathorWallet } from '@hathor/wallet-lib';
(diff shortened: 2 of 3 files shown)


packages/hathor-rpc-handler/src/errors/index.ts
```diff
@@ -18,3 +18,7 @@ export class DifferentNetworkError extends Error {};
 export class NoUtxosAvailableError extends Error {};
 
 export class SignMessageError extends Error {};
+
+export class ParamsValidationError extends Error {};
+
+export class BluePrintNotFoundError extends ParamsValidationError {};
```
pedroferreira1, line 24:
```suggestion
export class BlueprintNotFoundError extends ParamsValidationError {};
```

packages/hathor-rpc-handler/src/rpcMethods/sendNanoContractTx.ts
```diff
@@ -55,6 +61,20 @@ export async function sendNanoContractTx(
...
+      await ncApi.getBlueprintInformation(blueprint_id);
+    } catch (error) {
+      if (error instanceof NanoRequest404Error) {
+        throw new BluePrintNotFoundError();
+      }
+
+      throw new ParamsValidationError();
+    }
+  }
+
```
pedroferreira1, line 77: We should also check that the `nc_id` (if not executing a `initialize` method) is a valid nano_contract_id

alexruzenhack: todo: I believe it is lacking a validation on the methods existence after confirming that the blueprint exists.

alexruzenhack (changes requested):
It is lacking a validation. See this comment https://github.com/HathorNetwork/hathor-rpc-lib/pull/16#discussion_r1727407723

PR #18, opened 2024-08-26
docs(API): create reference documentation for API using Open RPC standard

BrunoCampana:
Objective:
- Document the API exposed by wallet applications to communicate with dApps on the Hathor platform.
- API type: JSON-RPC 2.0.
- API documented using the Open RPC standard.
- API documented in the `openrpc.json` file.
- The file should be used to automatically generate documentation on the Hathor docs website, similar to how it is already done with the full node and headless APIs.
- The final format that will be used is still being defined.
- For now, the following format can be used as a minimum standard to be used, but it should still improve the UI/UX:

https://beta-test.docs.hathor.network/references/dapps/rpc-api/htr_getBalance

Only one file to be reviewed: `openrpc.json`

### DIFF
--- a/docs/openrpc.json
+++ b/docs/openrpc.json
@@ -211,7 +211,7 @@
         },
         {
           "name": "maximumAmount",
-          "summary": "Returns only UTXOs smaller or equal than specified amount",
+          "summary": "Limit the maximum total amount to return summing all utxos",
           "schema": {
             "type": "number"
           }
@@ -378,7 +378,7 @@
     {
       "name": "htr_createToken",
       "summary": "Requests user to create token",
-      "description": "Provides user the deailts for a token creation. dApp provides wallet application with data that comprises a token creation. Wallet application assembles and signs transaction, and then pushes it to network or returns it to dApp.",
+      "description": "Provides user the details for a token creation. dApp provides wallet application with data that comprises a token creation. Wallet application assembles and signs transaction, and then pushes it to network or returns it to dApp.",
       "paramStructure": "by-name",
       "params": [
         {
@@ -580,13 +580,12 @@
       "Network": {
         "enum": [
           "testnet",
-          "mainnet"
+          "mainnet",
+          "privatenet"
         ]
       },
       "Token": {
-        "description": "32 bytes encoded in hexadecimal.",
-        "type": "string",
-        "pattern": "^[a-fA-F0-9]{64}$"
+        "$ref": "#/components/schemas/TransactionId"
       },
       "GetBalanceObject": {
         "type": "object",
@@ -615,9 +614,7 @@
         "type": "object",
         "properties": {
           "id": {
-            "description": "32 bytes encoded in hexadecimal.",
-            "type": "string",
-            "pattern": "^[a-fA-F0-9]{64}$"
+            "$ref": "#/components/schemas/Token"
           },
           "name": {
             "description": "UTF-8 characters.",
@@ -676,7 +673,7 @@
         }
       },
       "Address": {
-        "description": "20 bytes encoded in base58check.",
+        "description": "25 bytes encoded in base58check.",
         "type": "string",
         "pattern": "^[123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz]{34}$"
       },
@@ -710,12 +707,10 @@
             "$ref": "#/components/schemas/Address"
           },
           "amount": {
-            "type": "number"
+            "$ref": "#/components/schemas/OutputValueType"
           },
           "tx_id": {
-            "description": "32 bytes encoded in hexadecimal.",
-            "type": "string",
-            "pattern": "pattern": "^[a-fA-F0-9]{64}$"
+            "$ref": "#/components/schemas/TransactionId"
           },
           "locked": {
             "type": "boolean"
@@ -771,6 +766,11 @@
         "description": "Transaction encoded in hexadecimal.",
         "type": "string"
       },
+      "TransactionId": {
+        "description": "32 bytes encoded in hexadecimal; hash of transaction.",
+        "type": "string",
+        "pattern": "^[a-fA-F0-9]{64}$"
+      },
       "NanoContract": {
         "type": "object",
         "allOf": [
@@ -780,6 +780,7 @@
           {
             "properties": {
               "id": {
+                "description": "Refers to blueprint_id when method is initialize; refers to nc_id otherwise.",
                 "type": "string"
               },
               "method": {
@@ -844,7 +845,7 @@
             }
           },
           "hash": {
-            "type": "string"
+            "$ref": "#/components/schemas/TransactionId"
           }
         }
       },
@@ -852,8 +853,8 @@
         "type": "object",
         "properties": {
           "hash": {
-            "title": "Hash of transaction being spent",
-            "type": "string"
+            "title": "Hash of transaction's UTXO being spent",
+            "$ref": "#/components/schemas/TransactionId"
           },
           "index": {
             "title": "Index of outputs array from output being spent",
@@ -881,7 +882,7 @@
       },
       "Buffer": {
         "type": "string"
-      }
+      },
     },
     "errors":{
       "NotImplementedError": {
--- a/docs/openrpc.json
+++ b/docs/openrpc.json
@@ -20,7 +20,7 @@
       "params": [],
       "result": {
         "name": "ConnectedNetwork",
-        "summary": "Object containing network name and genesis hash",
+        "summary": "Provides network name and genesis hash",
         "schema": {
           "type": "object",
           "properties": {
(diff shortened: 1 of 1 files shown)


andreabadesso (approved):

docs/openrpc.json
```diff
@@ -0,0 +1,917 @@
...
+    "description": "API exposed by wallet applications for integration with dApps on Hathor platform.",
+    "contact": {
+      "name": "Hathor Labs",
+      "url": "https://docs.hathor.network/references/besides-documentation#support"
+    },
+    "license": {
+      "name": "MIT",
+      "url": "https://github.com/HathorNetwork/hathor-rpc-lib/blob/release/LICENSE"
+    },
+    "version": "0.1.0"
```
pedroferreira1, line 14: Is this the package version? If it is, it should be `0.0.2-experimental-alpha`

docs/openrpc.json
```diff
@@ -0,0 +1,917 @@
...
+        "name": "token",
+        "required": true,
+        "schema": {
+          "$ref": "#/components/schemas/Token"
+        }
+      }
+    },
+    "schemas": {
+      "Network": {
+        "enum": [
```
pedroferreira1, line 581: I think it can also be `privatenet`. We can confirm with @andreabadesso

docs/openrpc.json
```diff
@@ -0,0 +1,917 @@
...
+          }
+        }
+      },
+      "TokenInfo": {
+        "type": "object",
+        "properties": {
+          "id": {
+            "description": "32 bytes encoded in hexadecimal.",
+            "type": "string",
+            "pattern": "^[a-fA-F0-9]{64}$"
```
pedroferreira1, line 620: Can you use the `Token` schema?

docs/openrpc.json
```diff
@@ -0,0 +1,917 @@
...
+      }
+    },
+    "schemas": {
+      "Network": {
+        "enum": [
+          "testnet",
+          "mainnet"
+        ]
+      },
+      "Token": {
```
pedroferreira1, line 586: Maybe this schema name can be `TransactionId` or `TransactionHash`, so it can reused. What do you think?

docs/openrpc.json
```diff
@@ -0,0 +1,917 @@
...
+            "type": "boolean"
+          },
+          "melt": {
+            "title": "if has melt authority",
+            "type": "boolean"
+          }
+        }
+      },
+      "Address": {
+        "description": "20 bytes encoded in base58check.",
```
pedroferreira1, line 679: It's 25 bytes

docs/openrpc.json
```diff
@@ -0,0 +1,917 @@
...
+        }
+      },
+      "UtxoInfo": {
+        "type": "object",
+        "properties": {
+          "address": {
+            "$ref": "#/components/schemas/Address"
+          },
+          "amount": {
+            "type": "number"
```
pedroferreira1, line 713: OutputValue schema?

docs/openrpc.json
```diff
@@ -0,0 +1,917 @@
...
+        "type": "object",
+        "allOf": [
+          {
+            "$ref": "#/components/schemas/Transaction"
+          },
+          {
+            "properties": {
+              "id": {
+                "type": "string"
+              },
```
pedroferreira1, line 784: What's this id?

docs/openrpc.json
```diff
@@ -0,0 +1,917 @@
...
+            }
+          },
+          "tokens": {
+            "type": "array",
+            "items": {
+              "$ref": "#/components/schemas/Token"
+            }
+          },
+          "hash": {
+            "type": "string"
```
pedroferreira1, line 847: 32 bytes hexadecimal. If we change the `Token` schema as in my suggestion, we can use it here and other places like this

docs/openrpc.json
```diff
@@ -0,0 +1,917 @@
...
+          },
+          "data": {
+            "title": "Input signed data for P2PKH and redeemScript for P2SH",
+            "$ref": "#/components/schemas/Buffer"
+          }
+        }
+      },
+      "Output": {
+        "type": "object",
+        "properties": {
```
pedroferreira1, line 870: Does it have timelock?

docs/openrpc.json
```diff
@@ -0,0 +1,917 @@
...
+        {
+          "name": "amountBiggerThan",
+          "summary": "Selects only UTXOs greater than specified amount",
+          "schema": {
+            "type": "number"
+          }
+        },
+        {
+          "name": "maximumAmount",
+          "summary": "Returns only UTXOs smaller or equal than specified amount",
```
pedroferreira1, line 214: Is this correct? I think this is the maximum amount when you sum all utxos returned

docs/openrpc.json
```diff
@@ -0,0 +1,917 @@
...
+        },
+        {
+          "$ref": "#/components/errors/SendNanoContractTxFailure"
+        }
+      ]
+    },
+    {
+      "name": "htr_createToken",
+      "summary": "Requests user to create token",
+      "description": "Provides user the deailts for a token creation. dApp provides wallet application with data that comprises a token creation. Wallet application assembles and signs transaction, and then pushes it to network or returns it to dApp.",
```
pedroferreira1, line 381:
```suggestion
      "description": "Provides user the details for a token creation. dApp provides wallet application with data that comprises a token creation. Wallet application assembles and signs transaction, and then pushes it to network or returns it to dApp.",
```

docs/openrpc.json
```diff
@@ -0,0 +1,917 @@
...
+            "minLength": 1,
+            "maxLength": 5
+          }
+        },
+        {
+          "name": "amount",
+          "required": true,
+          "schema": {
+            "type": "number"
+          }
```
pedroferreira1, line 409: OutputValue schema?

docs/openrpc.json
BrunoCampana: Yes. This is the package version.

I suggest that the package versioning be changed:

From:
0.0.2-experimental-alpha

To:
0.1.1-experimental-alpha

Or:
0.2.0-experimental-alpha

(depending on whether the change between them was a new feature or a bug fix).

Let me explain:

- Following the semantic versioning standard https://semver.org , the first version released, even if still in the development phase, should be 0.1.0 and from then on increment the minor version for each subsequent release.

- In other words: the first version should start counting with minor and not with patch.

- According to: https://semver.org/#how-should-i-deal-with-revisions-in-the-0yz-initial-development-phase

docs/openrpc.json
BrunoCampana: OK: change made.

docs/openrpc.json
BrunoCampana: OK: change made: fixed error.

docs/openrpc.json
BrunoCampana: OK: change made: fixed typo.

docs/openrpc.json
BrunoCampana: OK: change made: this was indeed wrong. Fixed error.

docs/openrpc.json
BrunoCampana: Suggestion: keep as is:
Reason:
- In fact, these are values ​​that will always be the same, but they are different abstractions.
- The outputvalue refers specifically to the value of an output.
- Here, it refers to the number of tokens to be created.
- For example, this follows the same parameter description pattern that we already use in the headless "create token" API: https://docs.hathor.network/references/headless-wallet/http-api/create-token

docs/openrpc.json
BrunoCampana: OK: change made: replaced schema as suggested.

docs/openrpc.json
BrunoCampana: OK: change made:
- I didn't change the name.
- I think it's important to keep the token schema.
- Instead of replacing it, I added a new "TransactionId" schema and now the Token schema only references it.

docs/openrpc.json
BrunoCampana: OK: change made:
- I created a new `TransactionId` schema.
- I replaced it within the `Token` schema, here and everywhere else where TransactionId appears.

docs/openrpc.json
BrunoCampana: OK: change made:
- You are correct: the network can indeed be "privatenet", because it comes from wallet-lib, where we see this option in `hathor-wallet-lib/src/models/network.ts`

docs/openrpc.json
BrunoCampana: Suggestion: keep as is:
- An output only has:
- value
- token_data
- script
- Timelock is a property within the script.
- However, script here is encoded as base64, the script properties do not appear.

docs/openrpc.json
BrunoCampana: I'm not sure, but I believe `id` corresponds to:
- blueprint id when method is initiliaze.
- nc_id otherwise.

It seems to behave as follows:
https://github.com/HathorNetwork/hathor-wallet-lib/blob/master/src/nano_contracts/nano_contract.ts

And:
https://github.com/HathorNetwork/hathor-core/blob/feat/nano-sdk-v1/hathor/nanocontracts/nanocontract.py

I added the description to make it clear.

docs/openrpc.json
pedroferreira1: We can change it but in another PR. My suggestion is that we create a new PR to update to 0.1.0-experimental-alpha but this `docs.openrpc.json` should be the same as in the package.json. So I suggest you set as 0.0.1-experimental-alpha now and in the other PR we change it as well

docs/openrpc.json
pedroferreira1: They are not different abstractions, this amount is the amount of the output of the newly created token, so it's the same meaning.

docs/openrpc.json
BrunoCampana: OK: change made:
- Version changed to "0.0.3-experimental-alpha", which is the latest one in master.

docs/openrpc.json
BrunoCampana: The thing is, the schema you pointed out is for a parameter, not a property of the result. 
It's the parameter that defines how many tokens will be created, not the amount of tokens that were created. At this point, there is no output yet. There is only the data for the amount of tokens the DApp wants the wallet to create.

BrunoCampana: OK: change made:
- Replaced number by reference to `OutputValueType` schema.

pedroferreira1 (approved):

PR #30, opened 2025-02-12
feat: send tokens

andreabadesso:
### Acceptance Criteria

- We should be able to process the `sendTransaction` RPC request properly
- We should upgrade the wallet-lib to the latest version

### Checklist

### DIFF
--- a/packages/hathor-rpc-handler/src/rpcMethods/sendTransaction.ts
+++ b/packages/hathor-rpc-handler/src/rpcMethods/sendTransaction.ts
@@ -6,7 +6,8 @@
  */
 
 import { z } from 'zod';
-import type { HathorWallet } from '@hathor/wallet-lib';
+import type { HathorWallet, Transaction } from '@hathor/wallet-lib';
+import type { OutputSendTransaction } from '@hathor/wallet-lib/lib/wallet/types';
 import {
   TriggerTypes,
   PinConfirmationPrompt,
@@ -22,7 +23,12 @@ import {
   RpcResponse,
   RpcMethods,
 } from '../types';
-import { PromptRejectedError, InvalidParamsError } from '../errors';
+import {
+  PromptRejectedError,
+  InvalidParamsError,
+  SendTransactionError,
+  InsufficientFundsError,
+} from '../errors';
 import { validateNetwork } from '../helpers';
 
 const sendTransactionSchema = z.object({
@@ -31,7 +37,7 @@ const sendTransactionSchema = z.object({
     network: z.string().min(1),
     outputs: z.array(z.object({
       address: z.string().optional(),
-      value: z.number().positive(),
+      value: z.string().transform(val => BigInt(val)),
       token: z.string().optional(),
       type: z.string().optional(),
       data: z.array(z.string()).optional(),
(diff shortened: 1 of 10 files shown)


packages/hathor-rpc-handler/src/rpcMethods/sendTransaction.ts
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    return typedOutput;
+  });
+
+  // Create the transaction service but don't run it yet
+  const sendTransaction = await wallet.sendManyOutputsSendTransaction(sendTransactionOutputs, {
+    inputs: params.inputs || [],
+    changeAddress: params.changeAddress,
+    pinCode: '111111',
```
raul-oliveira, line 107: Magic string here. Assign a variable that holds the meaning of it. 

IE `stubPinCode` or `dumbPinCode`

raul-oliveira (approved):
NIT: change the pincode string to a more meaningful variable

packages/hathor-rpc-handler/src/rpcMethods/sendTransaction.ts
andreabadesso: Done! Thanks

packages/hathor-rpc-handler/src/rpcMethods/sendTransaction.ts
```diff
@@ -0,0 +1,188 @@
...
+} from '../errors';
+import { validateNetwork } from '../helpers';
+
+const sendTransactionSchema = z.object({
+  method: z.literal(RpcMethods.SendTransaction),
+  params: z.object({
+    network: z.string().min(1),
+    outputs: z.array(z.object({
+      address: z.string().optional(),
+      value: z.string().transform(val => BigInt(val)),
```
pedroferreira1, line 40: Isn't it optional when it's a data output?

packages/hathor-rpc-handler/src/rpcMethods/sendTransaction.ts
```diff
@@ -0,0 +1,188 @@
...
+      token: output.token,
+      type: output.type,
+      data: output.data,
+    };
+    const typedOutput = outputData as unknown as OutputSendTransaction;
+
+    if (typedOutput.type === 'data') {
+      (typedOutput as unknown as { value: bigint }).value = BigInt(1);
+      typedOutput.token = '00';
+    }
```
pedroferreira1, line 98: Do you need this? I think the lib method already handles it in case of a data output

packages/hathor-rpc-handler/src/rpcMethods/sendTransaction.ts
andreabadesso: It is, changed it to optional and added a better refine method to make it required only when not data output

packages/hathor-rpc-handler/src/rpcMethods/sendTransaction.ts
andreabadesso: removed it.

andreabadesso: We actually do need it because `prepareTxData` doesn't add the 1 HTR to data outputs.

pedroferreira1 (approved):

PR #31, opened 2025-03-13
feat: specific prepare tx error

andreabadesso:
### Motivation

We should have a specific error for when the transaction fails while being prepared, the idea is to differentiate it from the `SendTransactionError` (which happens after the transaction is displayed to the user) 

### Acceptance Criteria

- Throw a specific error for failures in the prepareTx phase

### Checklist


andreabadesso:
Moved to https://github.com/HathorNetwork/hathor-rpc-lib/pull/33

PR #32, opened 2025-03-14
fix: create token bigint

andreabadesso:
### Motivation

We should accept strings as the amount for the create token RPC and transform it to a BigInt before passing to the clients.

### Acceptance Criteria

- We should accept strings as the amount so we support large integers

### Checklist


andreabadesso:
PR moved to https://github.com/HathorNetwork/hathor-rpc-lib/pull/new/feat/bigint-support

PR #33, opened 2025-03-17
feat: bigint support

andreabadesso:
### Motivation

What was the motivation for the changes in this PR?

### Acceptance Criteria

- Include here all things that this PR should solve

### Checklist
- [ ] If you are requesting a merge into `master`, confirm this code is production-ready and can be included in future releases as soon as it gets merged
- [ ] Make sure either the unit tests and/or the QA tests are capable of testing the new features
- [ ] Make sure you do not include new dependencies in the project unless strictly necessary and do not include dev-dependencies as production ones. More dependencies increase the possibility of one of them being hijacked and affecting us.

### DIFF
--- a/packages/hathor-rpc-handler/src/rpcMethods/createToken.ts
+++ b/packages/hathor-rpc-handler/src/rpcMethods/createToken.ts
@@ -26,7 +26,7 @@ import { z } from 'zod';
 const createTokenSchema = z.object({
   name: z.string().min(1),
   symbol: z.string().min(1),
-  amount: z.number().positive(),
+  amount: z.string().transform(val => BigInt(val)),
   address: z.string().nullish().default(null),
   change_address: z.string().nullish().default(null),
   create_mint: z.boolean().default(true),
--- a/packages/hathor-rpc-handler/src/rpcMethods/createToken.ts
+++ b/packages/hathor-rpc-handler/src/rpcMethods/createToken.ts
@@ -26,7 +26,16 @@ import { z } from 'zod';
 const createTokenSchema = z.object({
   name: z.string().min(1),
   symbol: z.string().min(1),
-  amount: z.string().transform(val => BigInt(val)),
+  amount: z.string().refine(val => {
+    try {
+      const bigIntVal = BigInt(val);
+      return bigIntVal > BigInt(0);
+    } catch (e) {
+      return false;
+    }
+  }, {
+    message: "Amount must be a valid positive number string"
+  }).transform(val => BigInt(val)),
   address: z.string().nullish().default(null),
   change_address: z.string().nullish().default(null),
   create_mint: z.boolean().default(true),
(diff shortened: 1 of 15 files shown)


packages/hathor-rpc-handler/src/rpcMethods/createToken.ts
```diff
@@ -26,7 +26,16 @@ import { z } from 'zod';
...
+  amount: z.string().refine(val => {
+    try {
+      const bigIntVal = BigInt(val);
+      return bigIntVal > BigInt(0);
+    } catch (e) {
+      return false;
+    }
+  }, {
+    message: "Amount must be a valid positive number string"
+  }).transform(val => BigInt(val)),
```
r4mmer, line 38: suggestion: Use the coerce which will output a bigint.

```suggestion
  amount: z.union([z.bigint(), z.number(), z.string().regex(/^\d+$/)])
    .pipe(z.coerce.bigint().positive()),
```

Or if you want to change the error message with refine.
```ts
z.union([z.bigint(), z.number(), z.string().regex(/^\d+$/)])
  .pipe(z.coerce.bigint())
  .refine(val => val > 0n, "Amount must be a valid positive number.")
```

Obs: If you want to enable points and commas in the string version you will need to change the regex to something like `/^(\d[\.,]?)*\d$/`

packages/hathor-rpc-handler/src/rpcMethods/sendNanoContractTx.ts
```diff
@@ -24,11 +24,32 @@ import {
...
+      }
+    }, {
+      message: "Amount must be a valid non-negative number string"
+    })
+  ).transform(actions => 
+    actions.map(action => ({
+      ...action,
+      amount: BigInt(action.amount)
+    }))
+  ),
```
r4mmer, line 52: Maybe we should use the coercible bigint since it is less verbose.

r4mmer (changes requested):

packages/hathor-rpc-handler/src/rpcMethods/createToken.ts
andreabadesso: That's a great suggestion, thanks!

I refactored the code, but removed `z.bigint()` and `z.number()` as we will only accept number strings for now

packages/hathor-rpc-handler/src/rpcMethods/sendNanoContractTx.ts
andreabadesso: Done, thanks

pedroferreira1 (approved):

r4mmer (changes requested):

packages/hathor-rpc-handler/src/rpcMethods/sendNanoContractTx.ts
```diff
@@ -24,11 +24,22 @@ import {
...
+  amount: string,
+}
+
 const sendNanoContractSchema = z.object({
   method: z.string().min(1),
   blueprint_id: z.string().nullish(),
   nc_id: z.string().nullish(),
-  actions: z.array(z.custom<NanoContractAction>()),
+  actions: z.array(
+    z.custom<NanoContractActionWithStringAmount>().transform(action => ({
```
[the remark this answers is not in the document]
andreabadesso: Done, can you please check?

PR #34, opened 2025-03-20
refactor: we should expect one output for each data instead of an array of data outputs

andreabadesso:
### Acceptance Criteria

- We should expect a single data string in the outputs instead of an array of strings

### Checklist

### DIFF
--- a/packages/hathor-rpc-handler/__tests__/rpcMethods/sendTransaction.test.ts
+++ b/packages/hathor-rpc-handler/__tests__/rpcMethods/sendTransaction.test.ts
@@ -57,7 +57,7 @@ describe('sendTransaction', () => {
           inputs: [{
             txId: 'testTxId',
             index: 0,
-            value: 100,
+            value: 100n,
             address: 'testAddress',
             token: '00',
           }],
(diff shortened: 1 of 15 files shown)


pedroferreira1 (approved):

packages/hathor-rpc-handler/__tests__/rpcMethods/sendTransaction.test.ts
```diff
@@ -61,6 +61,11 @@ describe('sendTransaction', () => {
             address: 'testAddress',
             token: '00',
           }],
+          outputs: [{
+            address: 'testAddress',
+            value: BigInt(100),
```
r4mmer, line 66: Shouldn't all values be bigint? Should we change the input value as well?

r4mmer, line 66: Why not `100n`?

andreabadesso: Refactored everywhere to use `n` instead of `BigInt`

andreabadesso: Yes we should, done

r4mmer (approved):

PR #35, opened 2025-03-20
chore: updated openrpc with sendTransaction

andreabadesso:
### Motivation

The openrpc doc was outdated after the recent parameter changes

### Acceptance Criteria

- We should document the `htr_sendTransaction` RPC
- We should update the `value` and `amount` parameters for `htr_sendNanoContractTx` and `htr_createToken` as they're now strings.

### Checklist

### DIFF
--- a/docs/openrpc.json
+++ b/docs/openrpc.json
@@ -312,7 +312,7 @@
         {
           "name": "nc_id",
           "summary": "Id of contract being called",
-          "description": "Should be used for any method different than initialize, to execute a contract. Should not be used with method initialize."
+          "description": "Should be used for any method different than initialize, to execute a contract. Should not be used with method initialize.",
           "schema": {
             "description": "32 bytes encoded in hexadecimal.",
             "type": "string",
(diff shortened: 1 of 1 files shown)


BrunoCampana (approved):
Approved.
I have made improvements based on your changes.
In this case, the following:
- I improved the descriptions, to maintain standardization with the rest of the document and our documentation as a whole.
- Please note that several of the changes I made do NOT concern the additions that were made recently, but rather small improvements to the whole.
- I replaced the "title" members that I had previously used with "description" in everything. In the future, we will use this JSON file to generate automatic HTML documentation, as we already do for the core and headless HTTP APIs. Hence the importance of the adjustment.
- I replaced some types of "number" with "integer". In JSON, there are no integers and floats, this difference being made by the presence or absence of decimals or exponents. However, JSON Schema adds this and allows us to specify which type of "number" should be sent. This restriction is relevant at the documentation level to help inform the user-developer which values ​​he can send. I replaced "number" with "integer" in places where I was absolutely sure that they were "integers", i.e.: indexes. Since I haven't read all the code, much less wallet-lib, I preferred not to change all of them, but my guess is that all "numbers" in this API are "integer".
- I changed the description of `output` parameters in `sendTransaction`. I may be wrong, but from what I read in the method code, an `output` EITHER: has amount, value and token OR has type and data, when it is a data output. It is necessary to verify if this is correct.

pedroferreira1 (approved):

PR #36, opened 2025-03-31
refactor: push tx false in the htr_sendNanoContractTx should return a txHex

andreabadesso:
### Motivation

Currently, we're returning the entire `SendTransaction` object, which after being serialized by Reown, does not include the transaction hex. We should return only a string containing the `txHex` when `push_tx` is set to false.

Here is an example response (received on the dApp end) when passing `push_tx: false`:

```json
{
    "type": 0,
    "response": "0004000000013cb032600bdf7db784800e4ea911b10676fa2f67591f82bb62628c234e7715950a696e697469616c697a650024001976a9144ebb23975b8cbf06cd018dd4fd0216b759e3a3ac88ac000100000467eab8df2103961ed22bc0111ecbd9bc43bb67cbcf7155571b46c0fff5bc5769ec4910d6f583473045022100cfe9dc844e53a984976a266bc2010789f18dd39403c23daf5e7dce42ddc38975022036c8c470e4276f980b7e34971ce1b9e45e114076b1d4a0fdf744a559f2fd1b4f4030f50e7c7b57cb67eab8690000000000"
}
```

### Acceptance Criteria

- We should return the `txHex` when the `push_tx` parameter is set to `false`.

### Checklist

### DIFF
--- a/packages/hathor-rpc-handler/src/rpcMethods/sendNanoContractTx.ts
+++ b/packages/hathor-rpc-handler/src/rpcMethods/sendNanoContractTx.ts
@@ -6,7 +6,7 @@
  */
 
 import { z } from 'zod';
-import type { HathorWallet } from '@hathor/wallet-lib';
+import type { HathorWallet, SendTransaction } from '@hathor/wallet-lib';
 import {
   TriggerTypes,
   PinConfirmationPrompt,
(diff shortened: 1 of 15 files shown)


packages/hathor-rpc-handler/src/rpcMethods/sendNanoContractTx.ts
```diff
@@ -132,14 +128,36 @@ export async function sendNanoContractTx(
         args: confirmedArgs,
       };
 
-      const response = await sendMethod(
-        params.method,
-        caller, 
-        txData, 
-        {
-          pinCode: pinCodeResponse.data.pinCode,
+      let response: SendTransaction | string;
```
pedroferreira1, line 131: Isn't it `NanoContract | string`?

pedroferreira1 (approved):

PR #40, opened 2025-05-19
feat: added method for using nano headers in a create token tx

andreabadesso:
### Acceptance Criteria

- We should include the `htr_createNanoContractCreateTokenTx` rpc method
- We should use the schemas from the wallet-lib
- We should properly validate the token name and symbol sizes
- We should adapt existing methods to work with the wallet-lib

### Checklist

### DIFF
--- a/packages/hathor-rpc-handler/__tests__/rpcMethods/sendNanoContractTx.test.ts
+++ b/packages/hathor-rpc-handler/__tests__/rpcMethods/sendNanoContractTx.test.ts
@@ -315,32 +315,6 @@ describe('sendNanoContractTx', () => {
       type: TriggerTypes.SendNanoContractTxLoadingTrigger,
     }, {});
   });
-
-  it('should reject transactions with zero amount', async () => {
-    // Setup a request with zero amount
-    const zeroAmountAction = {
-      type: 'deposit',
-      address: 'test-address',
-      token: '00',
-      amount: '0',
-    } as NanoContractActionWithStringAmount;
-
-    const requestWithZeroAmount = {
-      ...rpcRequest,
-      params: {
-        ...rpcRequest.params,
-        actions: [zeroAmountAction] as unknown as NanoContractAction[],
-      },
-    } as SendNanoContractRpcRequest;
-
-    // The validation should fail with a specific error
-    await expect(
-      sendNanoContractTx(requestWithZeroAmount, wallet, {}, promptHandler)
-    ).rejects.toThrow(InvalidParamsError);
-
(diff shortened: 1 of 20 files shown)


packages/hathor-rpc-handler/__tests__/rpcMethods/sendNanoContractTx.test.ts
```diff
@@ -315,32 +315,6 @@ describe('sendNanoContractTx', () => {
...
-    } as SendNanoContractRpcRequest;
-
-    // The validation should fail with a specific error
-    await expect(
-      sendNanoContractTx(requestWithZeroAmount, wallet, {}, promptHandler)
-    ).rejects.toThrow(InvalidParamsError);
-
-    // Verify that the promptHandler wasn't called since validation should fail first
-    expect(promptHandler).not.toHaveBeenCalled();
-  });
```
andreabadesso, line 343: The lib's schema allows zero amount

packages/hathor-rpc-handler/src/rpcMethods/createNanoContractCreateTokenTx.ts
```diff
@@ -0,0 +1,152 @@
...
+    ncId: data?.nc_id ?? null,
+    actions: data?.actions ?? [],
+    method,
+    args: data?.args ?? [],
+    pushTx: push_tx,
+  };
+  // Only pass CreateTokenParams fields, fallback to null/empty for missing
+  const tokenParams: NanoContractCreateTokenParams = {
+    name: createTokenOptions?.name ?? '',
+    symbol: createTokenOptions?.symbol ?? '',
```
pedroferreira1, line 87: Why do you need this if these fields are required?

packages/hathor-rpc-handler/src/rpcMethods/createNanoContractCreateTokenTx.ts
```diff
@@ -0,0 +1,152 @@
...
+    amount: typeof createTokenOptions?.amount === 'string' ? BigInt(createTokenOptions.amount) : (createTokenOptions?.amount ?? 0n),
+    address: createTokenOptions?.address ?? null,
```
pedroferreira1, line 89: This parameter is `mintAddress`

packages/hathor-rpc-handler/src/rpcMethods/createNanoContractCreateTokenTx.ts
andreabadesso: To make typescript happy... They're already being validated using zod

r4mmer (approved):

PR #54, opened 2025-07-14
refactor: returning a flat list of balances in htr_getBalance

andreabadesso:
### Motivation

`wallet.getBalance('00');` returns an array with a single object containing the balance for the requested token, e.g.:

```
[
  {
    "token": {
      "id": "000003521effbc8efd7b746a118cdc7d41d7cc1bf9c5d1fa5de4f8453f14ba4f"
    },
    "balance": {
      "unlocked": "0",
      "locked": "0"
    },
    "transactions": 0,
    "lockExpires": null,
    "tokenAuthorities": {
      "unlocked": {
        "mint": "0",
        "melt": "0"
      },
      "locked": {
        "mint": "0",
        "melt": "0"
      }
    }
  }
]
```

We want to "flat" this array so we have a list of balances instead of a list of arrays of balances

### Acceptance Criteria

- We should flat the balances list
- We should fix the response from the get balances rpc to follow the same format as the other rpcs:
```
{
    type: TriggerResponseTypes.GetBalanceConfirmationResponse,
    data: true,
}
```

### Checklist

### DIFF
--- a/packages/hathor-rpc-handler/src/rpcMethods/getBalance.ts
+++ b/packages/hathor-rpc-handler/src/rpcMethods/getBalance.ts
@@ -47,7 +47,7 @@ export async function getBalance(
   promptHandler: TriggerHandler,
 ) {
   const parseResult = getBalanceSchema.safeParse(rpcRequest.params);
-  
+
   if (!parseResult.success) {
     throw new InvalidParamsError(parseResult.error.errors.map(e => `${e.path.join('.')}: ${e.message}`).join(', '));
   }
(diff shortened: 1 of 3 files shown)


pedroferreira1 (approved):

packages/hathor-rpc-handler/src/rpcMethods/getBalance.ts
```diff
@@ -60,19 +61,22 @@ export async function getBalance(
...
   };
 
-  const confirmed = await promptHandler(prompt, requestMetadata);
+  const confirmed = await promptHandler(
+    prompt,
+    requestMetadata
+  ) as GetBalanceConfirmationResponse;
 
-  if (!confirmed) {
+  if (!confirmed.data) {
```
tuliomir, line 79: question: Should this have the additional protection of an optional chaining operator? Can `confirmed` be `null` or `undefined`?
```suggestion
  if (!confirmed?.data) {
```

tuliomir (approved):

PR #56, opened 2025-07-17
feat: add rpc request params to the prompt handler

pedroferreira1:
### Motivation

For the metamask snaps screens, it's important to show the request parameters in the prompt handler, so the user can decide to accept/reject the request.

### Acceptance Criteria

- Add the request parameters to the prompt handler argument object.

### Checklist


packages/hathor-rpc-handler/src/rpcMethods/signWithAddress.ts
```diff
@@ -73,8 +73,8 @@ export async function signWithAddress(
   };
 
   const prompt: SignMessageWithAddressConfirmationPrompt = {
+    ...rpcRequest,
```
andreabadesso, line 76: Hey, I think you should pass the validated params (after zod) on all methods

Because in some cases we are not only validating but also transforming the data (e.g. string -> bigint)

PR #57, opened 2025-07-23
fix: get address request confirmation should have a type for it

pedroferreira1:
### Motivation

After a previous refactor, the get request address confirmation prompt should have a type for its return.

### Acceptance Criteria

- Add return type for `GetRequestAddressConfirmation` prompt

### Checklist

### DIFF
--- a/packages/hathor-rpc-handler/src/rpcMethods/getAddress.ts
+++ b/packages/hathor-rpc-handler/src/rpcMethods/getAddress.ts
@@ -9,6 +9,7 @@ import { z } from 'zod';
 import type { HathorWallet } from '@hathor/wallet-lib';
 import {
   AddressRequestClientResponse,
+  AddressRequestConfirmationResponse,
   TriggerTypes,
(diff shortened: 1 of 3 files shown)


andreabadesso (approved):

packages/hathor-rpc-handler/src/rpcMethods/getAddress.ts
```diff
@@ -106,9 +107,9 @@ export async function getAddress(
         data: {
           address,
         }
-      }, requestMetadata);
+      }, requestMetadata) as AddressRequestConfirmationResponse;
 
-      if (!confirmed) {
+      if (!confirmed.data) {
```
tuliomir, line 112: suggestion: Since this is a library request, I'd suggest protecting it against a `TypeError` and always return the custom Error class.
```suggestion
      if (!confirmed?.data) {
```

tuliomir (approved):

PR #58, opened 2025-07-23
feat: add rpc request handlers with dialogs

pedroferreira1:
### Acceptance Criteria

- Add dialogs and prompt handlers for all RPC requests
- Add utils for changing network and start the wallet with the persisted network
- Use json stringify to handle rpc response

**Get address**

image

image

**Get balance**

image

image

**Get utxos**

image

image

**Send transaction**

image

image

**Create token**

image

image

**Create nano**

image

image

**Signed oracle data**

image

image

**Sign with address**

image

image

**Change network**

image

image

### Checklist

### DIFF
--- a/packages/snap/src/utils/wallet.ts
+++ b/packages/snap/src/utils/wallet.ts
@@ -1,5 +1,5 @@
 import { constants as libConstants, walletUtils, HathorWalletServiceWallet, Network } from '@hathor/wallet-lib';
-import { DEFAULT_PIN_CODE, REQUEST_METHODS, nodeURL, txMiningURL, walletServiceURL } from '../constants';
+import { DEFAULT_PIN_CODE, REQUEST_METHODS } from '../constants';
 import { getNetworkData, configNetwork } from './network';
 
 export const getHathorWallet = async (): HathorWalletServiceWallet => {


packages/snap/src/dialogs/sendTransaction.tsx
```diff
@@ -0,0 +1,106 @@
...
+      <Text><Bold>Change address</Bold></Text>
+      <Divider />
+      <Text>{changeAddress}</Text>
+    </Box>
+  );
+}
+
+export const sendTransactionPage = async (data, params, origin) => (
+  await snap.request({
+    method: 'snap_dialog',
```
andreabadesso, line 83: Use the constant here? `REQUEST_METHODS`

packages/snap/src/dialogs/signWithAddress.tsx
```diff
@@ -0,0 +1,29 @@
+import { Bold, Box, Copyable, Container, Heading, Section, Text } from '@metamask/snaps-sdk/jsx';
+
+export const signWithAddressPage = async (data, params, origin) => (
+  await snap.request({
+    method: 'snap_dialog',
+    params: {
+      type: 'confirmation',
```
andreabadesso, line 7: Use `DIALOG_TYPES.CONFIRMATION`?

packages/snap/src/utils/prompt.ts
```diff
@@ -1,20 +1,89 @@
...
+      approved = await addressPage(data, params, origin);
+      return { data: approved };
+    case TriggerTypes.GetUtxosConfirmationPrompt:
+      approved = await utxosPage(data, params, origin);
+      return { data: approved };
+    case TriggerTypes.ChangeNetworkConfirmationPrompt:
+      if (!(params.newNetwork in NETWORK_MAP)) {
+        // Reject if the newNetwork is not a valid option
+        return { data: false };
+      }
```
pedroferreira1, line 34: - Should I move this logic to another file? Maybe each rpc method should have a prompt handler file/method

I can change this in another PR if we decide it's better

packages/snap/src/dialogs/utxos.tsx
```diff
@@ -0,0 +1,96 @@
...
+    return libConstants.DEFAULT_NATIVE_TOKEN_CONFIG.symbol;
+  }
+
+  return '';
+}
+
+const renderAmountSummary = (data, params) => {
+  if (params.authorities) {
+    let authority = '';
+    if (params.authorities === 1) {
```
tuliomir, line 24: suggestion(non-blocking): Use the lib constants for those authority values
```suggestion
    if (params.authorities === libConstants.TOKEN_MINT_MASK) {
```

packages/snap/src/dialogs/utxos.tsx
```diff
@@ -0,0 +1,96 @@
...
+  );
+}
+
+export const utxosPage = async (data, params, origin) => {
+  const content = (
+    <Container backgroundColor='alternative'>
+      <Box>
+        <Heading>UTXO Details</Heading>
+        <Text>
+          {`${origin} requests access to ${data.utxos?.length || 0} UTXOs`}
```
tuliomir, line 46: suggestion(wording): I think "requests access" may be incorrectly interpreted by users as offering the funds of the UTXO to the origin dApp.

I'd suggest changing this to something like "${origin} requests information about ${n} UTXOs" or "${origin} requests data from ${n} UTXOs"

packages/snap/src/utils/network.ts
```diff
@@ -0,0 +1,47 @@
...
+/*
+ * Get persisted network data or the default from constants
+ */
+export const getNetworkData = async () => {
+  const persistedData = await snap.request({
+    method: REQUEST_METHODS.MANAGE_STATE,
+    params: { operation: 'get' },
+  }) ?? {};
+
+  const network = persistedData.network ?? DEFAULT_NETWORK;
```
tuliomir, line 13: question: Maybe this is out of the scope of this feature, but: Is this the full network name?

For example, if the request tries to change between `testnet` and `nano-testnet` will this show any difference between the two?

packages/snap/src/utils/prompt.ts
```diff
@@ -1,20 +1,89 @@
...
+  let approved;
+  switch (promptRequest.type) {
+    case TriggerTypes.SignMessageWithAddressConfirmationPrompt:
+      approved = await signWithAddressPage(data, params, origin);
+      return { data: approved };
+    case TriggerTypes.PinConfirmationPrompt:
+      return {
+        data: {
+          accepted: true,
+          pinCode: '123',
```
tuliomir, line 18: question: Should we use a constant or config for this value, so that it is easier to identify in future refactorings?

packages/snap/src/dialogs/address.tsx
```diff
@@ -1,20 +1,80 @@
-import { Box, Text } from '@metamask/snaps-sdk/jsx';
 import { REQUEST_METHODS, DIALOG_TYPES } from '../constants';
+import { Box, Container, Copyable, Heading, Section, Text } from '@metamask/snaps-sdk/jsx';
 
-export const addressPage = async () => (
+const renderParamText = (params) => {
+  switch (params.type) {
+    case 'first_empty':
```
tuliomir, line 6: question(non-blocking): Does it make sense at this moment in the snaps development to extract those parameter types to constants?

packages/snap/src/utils/prompt.ts
tuliomir: It doesn't seem necessary to me: the logic is still simple enough to stay in the same file.

packages/snap/src/utils/network.ts
pedroferreira1: Yes, it's different because I have different network URLs for each of these networks. It depends on what I want to support in the constants file.

```
export const NETWORK_MAP = {
  'mainnet': MAINNET_URLS,
  'nano-bravo-testnet': NANO_BRAVO_URLS,
  'testnet': TESTNET_URLS,
  'testnet-hotel': TESTNET_URLS,
}
```

packages/snap/src/dialogs/address.tsx
pedroferreira1: I think this ENUM should exist in the RPC handler package, where the rpc method definition is. I created an issue to handle this https://github.com/HathorNetwork/hathor-rpc-lib/issues/61

packages/snap/src/dialogs/sendTransaction.tsx
pedroferreira1: Done, thanks! https://github.com/HathorNetwork/hathor-rpc-lib/pull/58/commits/6eed62900560993d3bce7ea2754232a6343df537

packages/snap/src/dialogs/signWithAddress.tsx
pedroferreira1: Done, thanks! https://github.com/HathorNetwork/hathor-rpc-lib/pull/58/commits/6eed62900560993d3bce7ea2754232a6343df537

packages/snap/src/dialogs/address.tsx
```diff
@@ -1,20 +1,80 @@
-import { Box, Text } from '@metamask/snaps-sdk/jsx';
```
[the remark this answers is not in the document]
pedroferreira1: I will create a new PR to add the header in all files

packages/snap/src/dialogs/utxos.tsx
pedroferreira1: It makes sense, it's better like you suggested. https://github.com/HathorNetwork/hathor-rpc-lib/pull/58/commits/5f5a0f61512dcd7940e2d6bfc0f4c759f9af9015

tuliomir (approved):

PR #98, opened 2025-10-30
feat: v2

andreabadesso:
Application can be previewed at:
https://preview-webwallet-testt-khiynw-ksiija.webwallet.hathor.dev/

https://github.com/user-attachments/assets/5c041d7a-6ea0-4b4a-80ce-e9787cf5ca2a

You must be running the Snap on port 8080

### Acceptance Criteria
- Read-only wallet by requesting the `xpubkey` from the MetaMask Snap
- Connect/Disconnect Wallet - Connect MetaMask Snap to access wallet, persist session, disconnect to clear data
- Switch Networks - Switch between mainnet and testnet with automatic rollback on failure
- View Balances - Display `HTR` and custom token balances (available + locked)
- Manage Address - View current address with copy-to-clipboard
- Send Transactions - Send HTR or tokens to any address with optional timelock
- View Transaction History - Browse paginated transaction history for HTR or any token
- Register Custom Tokens - Add tokens via configuration string with automatic NFT detection
- Filter Tokens - Filter token list by Tokens/NFTs
- Real-time Updates - Auto-refresh balances and transaction list via WebSocket
- Error Recovery - Handle snap crashes, network errors, and connection issues with clear recovery paths.

### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,6 @@ dist/
 build/
 coverage/
 .cache/
-public/
 *.swp
 *.swo
 .claude


packages/web-wallet/README.md
```diff
@@ -0,0 +1,139 @@
...
+## Quick Start
+
+```bash
+# Install dependencies (from monorepo root)
+yarn install
+
+# Start development server
+yarn workspace @hathor/web-wallet dev
+
+# Access at http://localhost:5173
```
r4mmer: Tried and MetaMask said loading local snaps is disabled, can we also have a small segment on any configuration that we need to do on the MetaMask to make it work?

packages/web-wallet/nginx.conf
```diff
@@ -0,0 +1,26 @@
+server {
+    listen 80;
+    server_name localhost;
```
r4mmer:
```suggestion
    listen 80 default_server;
```

packages/web-wallet/playwright.config.ts
```diff
@@ -0,0 +1,42 @@
...
+    // Firefox and Webkit disabled - browsers not installed
+    // Uncomment and run `npx playwright install` to enable
+    // {
+    //   name: 'firefox',
+    //   use: { ...devices['Desktop Firefox'] },
+    // },
+    // {
+    //   name: 'webkit',
+    //   use: { ...devices['Desktop Safari'] },
+    // },
```
r4mmer: Should we test support for firefox and safari?

packages/web-wallet/playwright.config.ts
```diff
@@ -0,0 +1,42 @@
...
+  },
+
+  projects: [
+    {
+      name: 'chromium',
+      use: {
+        ...devices['Desktop Chrome'],
+        // Add extension support when available
+        // args: ['--load-extension=./extensions/metamask-flask'],
+        // headless: false,
```
r4mmer: Should we add these dependencies or not?

packages/web-wallet/src/App.tsx
```diff
@@ -0,0 +1,25 @@
+import { BrowserRouter } from 'react-router-dom'
+import { MetaMaskProvider } from '@hathor/snap-utils'
+import { WalletProvider } from './contexts/WalletContext'
+import WalletHome from './components/WalletHome'
+import { ErrorBoundary } from './components/ErrorBoundary'
+
+function App() {
+  return (
+    <ErrorBoundary>
```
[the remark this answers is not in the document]
r4mmer: On the issue of error boundaries maybe we should implement something more like [what is on the react docs](https://react.dev/reference/react/Component#catching-rendering-errors-with-an-error-boundary).

Also from the same page is [this](https://react.dev/reference/react/Component#componentdidcatch) where they recommend using [this dependency](https://github.com/bvaughn/react-error-boundary).

packages/web-wallet/src/contexts/WalletContext.tsx
```diff
@@ -0,0 +1,235 @@
...
+      // We could expose setters if needed
+    },
+    onError: setError,
+    onForceDisconnect: () => {
+      connection.disconnectWallet();
+    },
+  });
+
+  // Aggregate state from all hooks
+  const contextValue: WalletContextType = {
```
r4mmer: Since this is the heart of the wallet context I think we should expand on the explanation of each state being managed.

For instance, we can use `setError` method to set the value of `error` state but what is the purpose of this error management?
We already have multiple error boundaries on the code, why and where will this one be used?

packages/web-wallet/src/components/ui/dialog.tsx
```diff
@@ -0,0 +1,120 @@
...
+}: React.HTMLAttributes<HTMLDivElement>) => (
+  <div
+    className={cn(
+      "flex flex-col space-y-1.5 text-center sm:text-left",
+      className
+    )}
+    {...props}
+  />
+)
+DialogHeader.displayName = "DialogHeader"
```
r4mmer: display name? Will this be displayed to the user or on logs/renders?

packages/web-wallet/src/components/DisconnectConfirmModal.tsx
```diff
@@ -0,0 +1,73 @@
...
+            </p>
+            <p className="text-sm">
+              All wallet data stored locally will be permanently cleared, including:
+            </p>
+            <ul className="list-disc list-inside space-y-1 text-sm ml-1">
+              <li>Wallet connection information</li>
+              <li>Network preferences</li>
+              <li>Registered tokens</li>
+              <li>Address mode settings</li>
+            </ul>
```
r4mmer: This list is very important, maybe we should document these locally stored data and where to update the UI when they change.

packages/web-wallet/src/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  componentDidCatch(error: Error, errorInfo: React.ErrorInfo) {
+    console.error('ErrorBoundary caught an error:', error, errorInfo);
+  }
+
+  render() {
+    if (this.state.hasError) {
+      if (this.props.fallback) {
+        return this.props.fallback;
+      }
```
r4mmer: So we support fallback component and a default error component?

packages/web-wallet/src/components/Header.tsx
```diff
@@ -0,0 +1,249 @@
...
+                )}
+              </button>
+
+              {/* Desktop: Compact dropdown - only on desktop, stays in relative container */}
+              {isMenuOpen && (
+                <div className="hidden md:block absolute top-full right-0 mt-2 w-48 bg-[#191C21] border border-[#24292F] rounded-lg shadow-lg overflow-hidden z-50">
+                  <button
+                    onClick={() => {
+                      setIsMenuOpen(false);
+                      onCreateTokenClick?.();
```
r4mmer: I have never seen optional chaining operator used on functions, can we be sure this works?

packages/web-wallet/src/components/Header.tsx
```diff
@@ -0,0 +1,249 @@
...
+              >
+                {isMenuOpen ? (
+                  <X className="w-4 h-4 md:w-5 md:h-5 text-white" />
+                ) : (
+                  <Menu className="w-4 h-4 md:w-5 md:h-5 text-white" />
+                )}
+              </button>
+
+              {/* Desktop: Compact dropdown - only on desktop, stays in relative container */}
+              {isMenuOpen && (
```
r4mmer: I understand that there is an implementation for desktop and mobile.
But since the structure is so similar maybe this should be a separate component to avoid repetition.
Specially in this case where we can develop on desktop and forget to update the mobile implementation.

packages/web-wallet/src/components/ReceiveDialog.tsx
```diff
@@ -0,0 +1,110 @@
...
+    if (!displayAddress) return;
+
+    try {
+      await navigator.clipboard.writeText(displayAddress);
+      setCopied(true);
+      setTimeout(() => setCopied(false), 2000);
+    } catch (err) {
+      console.error('Failed to copy:', err);
+    }
+  };
```
r4mmer: The setTimeout not being tracked may cause issues

packages/web-wallet/src/components/UnregisterTokenDialog.tsx
```diff
@@ -0,0 +1,137 @@
...
+  onConfirm,
+  tokenSymbol,
+  tokenName,
+}) => {
+  const [isLoading, setIsLoading] = useState(false);
+  const [isConfirmed, setIsConfirmed] = useState(false);
+  const [error, setError] = useState<string | null>(null);
+  const [success, setSuccess] = useState(false);
+
+  const handleUnregister = async () => {
```
r4mmer: On the register token dialog we used `const { registerToken } = useWallet();` to register the token.
Why do we not have a similar method here for unregistering the token?

From the args it seems we receive the token and only notify the caller that the user intends to unregister and any actual action should be taken from the caller.

packages/web-wallet/src/components/WalletHome.tsx
```diff
@@ -0,0 +1,312 @@
...
+                  className="w-5 h-5 md:w-6 md:h-6"
+                />
+                <span className="text-xl md:text-2xl font-medium text-white">
+                  {balances.length > 0 ? `${formatHTRAmount(balances[0].available, false)} HTR` : '0 HTR'}
+                </span>
+              </div>
+              {/* Show custom token and NFT counts */}
+              {(customTokenCount > 0 || nftCount > 0) && (
+                <p className="text-xs text-muted-foreground">
+                  You also have {customTokenCount > 0 && `${customTokenCount} custom token${customTokenCount > 1 ? 's' : ''}`}
```
r4mmer: question: Are we planning any i18n in the future?

packages/web-wallet/src/contexts/WalletContext.tsx
```diff
@@ -0,0 +1,235 @@
...
+    address: balance.address || connection.address,
+    balances: balance.balances.length > 0 ? balance.balances : connection.balances,
+
+    // Token state
+    registeredTokens: tokens.registeredTokens.length > 0 ? tokens.registeredTokens : connection.registeredTokens,
+    selectedTokenFilter: tokens.selectedTokenFilter,
+    selectedTokenForSend: null, // This was removed in refactor
+
+    // Transaction state
+    isHistoryDialogOpen: transactions.isHistoryDialogOpen,
```
r4mmer: Should this be part of the wallet state?

packages/web-wallet/src/components/WalletHome.tsx
```diff
@@ -0,0 +1,312 @@
...
+      />
+      <AddressModeDialog
+        isOpen={addressModeOpen}
+        onClose={closeDialog}
+      />
+
+      {/* Error notification */}
+      {error && (
+        <ErrorNotification
+          error={new Error(error)}
```
r4mmer: We convert to `Error` just to use the `error.message`? 
Maybe we should just use the error string itself

packages/web-wallet/src/components/WalletHome.tsx
```diff
@@ -0,0 +1,312 @@
...
+          <TransactionNotification
+            transaction={notification}
+            onDismiss={clearNewTransaction}
+            onViewHistory={() => {
+              clearNewTransaction();
+              navigate('?dialog=history');
+            }}
+          />
+        );
+      })()}
```
r4mmer: Nice touch to define a one-use logic and avoid creating many one-off functions.

packages/web-wallet/src/components/WalletHome.tsx
```diff
@@ -0,0 +1,312 @@
...
+
+      {/* Transaction notification */}
+      {(() => {
+        if (!newTransaction) return null;
+        const tx = newTransaction as Record<string, unknown>;
+        if (tx.tx_id) return null; // This is for history dialog, not notification
+
+        const notification = tx as { type: 'sent' | 'received'; amount: bigint; timestamp: number };
+        return (
+          <TransactionNotification
```
r4mmer: What happens if we receive 2 transactions close to each other?

packages/web-wallet/src/components/__tests__/SendDialog.test.tsx
```diff
@@ -0,0 +1,94 @@
+import { describe, it, expect } from 'vitest';
```
r4mmer: This file does not test anything from the `SendDialog` but at least it is testing something from outside the test, but why test something from outside the repository?

packages/web-wallet/src/components/ui/button.tsx
```diff
@@ -0,0 +1,57 @@
...
+    return (
+      <Comp
+        className={cn(buttonVariants({ variant, size, className }))}
+        ref={ref}
+        {...props}
+      />
+    )
+  }
+)
+Button.displayName = "Button"
```
r4mmer: Why?

packages/web-wallet/src/constants.ts
```diff
@@ -0,0 +1,66 @@
...
+// Network configuration
+export const NETWORKS = {
+  MAINNET: 'mainnet',
+  TESTNET: 'testnet'
+} as const;
+
+// WafpWYepbV13FVM9Qp9brmBTXgjrn3dnfx 
+// Hathor API endpoints
+export const HATHOR_API_URLS = {
+  MAINNET: 'https://node.hathor.network/v1a',
```
r4mmer: This is not a valid node

packages/web-wallet/src/constants.ts
```diff
@@ -0,0 +1,66 @@
...
+  TESTNET: 'https://node1.india.testnet.hathor.network/v1a'
+} as const;
+
+// Hathor explorer URLs
+export const HATHOR_EXPLORER_URLS = {
+  MAINNET: 'https://explorer.hathor.network',
+  TESTNET: 'https://explorer.testnet.hathor.network'
```
r4mmer: This explorer will not connect to the same network as the testnet node

packages/web-wallet/src/constants.ts
```diff
@@ -0,0 +1,66 @@
...
+
+// UI Constants
+export const QR_CODE_SIZE = 200;
+export const TRANSACTION_HISTORY_LIMIT = 50;
+
+// Snap configuration (these should match the snap's configuration)
+export const SNAP_CONFIG = {
+  DEFAULT_PORT: 8080,
+  LOCAL_ORIGIN: `local:http://localhost:8080`
+} as const;
```
r4mmer: Is this right?

packages/web-wallet/src/contexts/__tests__/WalletContext.NetworkChange.test.tsx
```diff
@@ -0,0 +1,313 @@
...
+      const originalError = 'Network RPC failed: timeout';
+      const rollbackSucceeded = true;
+
+      const finalError = rollbackSucceeded
+        ? `Failed to change network: ${originalError}. Reverted to testnet.`
+        : `Network change failed (${originalError}) and rollback also failed. Please reconnect your wallet.`;
+
+      expect(finalError).toContain(originalError);
+      expect(finalError).toContain('Reverted to testnet');
+    });
```
r4mmer: What are we testing exatcly?

packages/web-wallet/src/contexts/hooks/useAddressMode.ts
```diff
@@ -0,0 +1,37 @@
...
+    // Refresh the displayed address to reflect the new mode
+    if (readOnlyWalletService.isReady()) {
+      try {
+        const address = await getDisplayAddressForMode(mode, readOnlyWalletService);
+        onAddressUpdate(address);
+      } catch (error) {
+        console.error('Failed to refresh address after mode change:', error);
+      }
+    }
+  };
```
r4mmer: Why can't the wallet context handle this?

packages/web-wallet/src/contexts/WalletContext.tsx
```diff
@@ -0,0 +1,235 @@
...
+      connection.setNetwork(network);
+      connection.setAddress(address);
+      connection.setBalances(balances);
+      connection.setRegisteredTokens(newTokens);
+      setError(warning);
+    },
+    onLoadingChange: () => {
+      // Loading state is managed internally by connection hook
+      // We could expose setters if needed
+    },
```
r4mmer: Same here

packages/web-wallet/src/contexts/hooks/useNetworkManagement.ts
```diff
@@ -0,0 +1,238 @@
...
+      // Load registered tokens for new network
+      const genesisHash = '';
+      const tokenLoadResult = await loadTokensWithBalances(newNetwork, genesisHash, {
+        detailedErrors: false,
+      });
+
+      // Update localStorage with new network
+      localStorage.setItem(STORAGE_KEYS.NETWORK, newNetwork);
+
+      onNetworkChange({
```
r4mmer: Should we await this?

packages/web-wallet/src/contexts/hooks/useWalletBalance.ts
```diff
@@ -0,0 +1,52 @@
...
+  const refreshBalance = useCallback(async () => {
+    if (!isConnected || !readOnlyWalletService.isReady()) return;
+
+    try {
+      const newBalances = await readOnlyWalletService.getBalance(TOKEN_IDS.HTR);
+
+      // Also refresh token balances
+      await onRefreshTokenBalances();
+
+      setBalances(newBalances);
```
r4mmer: Are we only refreshing the HTR balance?

packages/web-wallet/src/contexts/hooks/useWalletConnection.ts
```diff
@@ -0,0 +1,502 @@
...
+        }
+        currentSnapNetwork = parsedNetworkTest?.response?.network;
+      } catch (testError) {
+        console.error(testError);
+        throw new Error('Snap is not responding. Please make sure it is installed correctly.');
+      }
+
+      const targetNetwork = DEFAULT_NETWORK;
+      if (currentSnapNetwork !== targetNetwork) {
+        setLoadingStep(`Changing snap network to ${DEFAULT_NETWORK}...`);
```
r4mmer: Wait, so the network is hardcoded always?

packages/web-wallet/src/contexts/hooks/useWalletConnection.ts
```diff
@@ -0,0 +1,502 @@
...
+          method: 'htr_getXpub',
+          params: {}
+        });
+        const xpubTimeout = new Promise((_, reject) =>
+          setTimeout(() => reject(new Error('Get xpub timeout')), SNAP_TIMEOUTS.RPC_CALL)
+        );
+        xpubResponse = await Promise.race([xpubPromise, xpubTimeout]);
+      } catch (snapError) {
+        throw new Error(`Failed to get xpub: ${snapError instanceof Error ? snapError.message : String(snapError)}`);
+      }
```
r4mmer: We have a lot of races that could be better structured

packages/web-wallet/src/contexts/hooks/useWalletConnection.ts
```diff
@@ -0,0 +1,502 @@
...
+        walletAddress = await getDisplayAddressForMode(addressMode, readOnlyWalletService);
+      } catch {
+        throw new Error('Failed to retrieve wallet address. The wallet may not be properly initialized.');
+      }
+
+      setLoadingStep('Loading balance...');
+
+      const walletBalances = await readOnlyWalletService.getBalance(TOKEN_IDS.HTR);
+
+      const genesisHash = '';
```
r4mmer: Again the genesisHash is hardcoded empty

packages/web-wallet/src/contexts/hooks/useWalletConnection.ts
```diff
@@ -0,0 +1,502 @@
...
+    }
+  };
+
+  const disconnectWallet = () => {
+    const keysToRemove = Object.keys(localStorage).filter(key =>
+      key.startsWith('hathor')
+    );
+    keysToRemove.forEach(key => localStorage.removeItem(key));
+
+    readOnlyWalletService.stop();
```
r4mmer: From what I remember we did not setup event listeners or websocket, so what is there to stop on the lib?
Would just deleting the object not clear everything?

packages/web-wallet/src/hooks/useTokenHistory.ts
```diff
@@ -0,0 +1,59 @@
...
+  const refresh = useCallback(() => {
+    setHasMore(true);
+    return loadHistory(10, 0);
+  }, [loadHistory]);
+
+  const loadMore = useCallback(() => {
+    if (!isLoading && hasMore) {
+      return loadHistory(10, history.length);
+    }
+  }, [isLoading, hasMore, history.length, loadHistory]);
```
r4mmer: Will this work with over 20 txs in the history?
If the idea is "load 10 after the `history.length`" it may not work since we get 10 txs and use `setHistory` with it meaning the history.length will always be 10 if we have more than 20 txs and we will always load the same 11 to 20

packages/web-wallet/src/hooks/useTokens.ts
```diff
@@ -0,0 +1,82 @@
...
+      uid: TOKEN_IDS.HTR,
+      name: HTR_TOKEN_INFO.name,
+      symbol: HTR_TOKEN_INFO.symbol,
+      balance: {
+        available: htrBalance.available,
+        locked: htrBalance.locked,
+      },
+      isNFT: false,
+      registeredAt: 0, // HTR is always available
+    };
```
r4mmer: An idea would be to have the token data and metadata separate.
This would mean that token data list would be more stable and would change less often.
Metadata may change as much as needed but will not require updates on other stuff.

packages/web-wallet/src/hooks/useTokens.ts
```diff
@@ -0,0 +1,82 @@
...
+  }, [allTokens]);
+
+  const nftCount = useMemo(() => {
+    return allTokens.filter(t => t.isNFT).length;
+  }, [allTokens]);
+
+  // Calculate custom token count (excluding HTR)
+  const customTokenCount = useMemo(() => {
+    return registeredTokens.filter(t => !t.isNFT).length;
+  }, [registeredTokens]);
```
r4mmer: Why this count?

packages/web-wallet/src/services/ReadOnlyWalletService.ts
```diff
@@ -0,0 +1,402 @@
...
+        console.log('New transaction received:', tx.tx_id);
+      });
+
+      this.wallet.on('update-tx', (tx: { tx_id: string }) => {
+        console.log('Transaction updated:', tx.tx_id);
+      });
+
+      this.wallet.on('reload-data', () => {
+        console.log('Connection restored, data should be reloaded');
+      });
```
r4mmer: Maybe move this to debug?

packages/web-wallet/src/services/ReadOnlyWalletService.ts
```diff
@@ -0,0 +1,402 @@
...
+        } else {
+          // For other errors, re-throw
+          throw error;
+        }
+      }
+
+      console.log('Read-only wallet ready');
+    } catch (error) {
+      console.error('Failed to initialize read-only wallet:', error);
+      this.isInitializing = false; // Reset flag before clearing wallet
```
r4mmer: Already set on finally block

packages/web-wallet/src/services/ReadOnlyWalletService.ts
```diff
@@ -0,0 +1,402 @@
...
+  address: string;
+  index: number;
+  transactions: number;
+}
+
+/**
+ * Service for managing a read-only wallet instance using wallet-lib.
+ * This wallet can perform all read operations but cannot sign transactions.
+ */
+export class ReadOnlyWalletService {
```
r4mmer: `ReadOnlyWalletWrapper` seems more indicative of what this is.

packages/web-wallet/src/services/TokenRegistryService.ts
```diff
@@ -0,0 +1,293 @@
...
+    // Check cache first
+    const cached = this.tokenCache.get(tokenUid);
+    if (cached) {
+      return cached;
+    }
+
+    // Fetch from wallet-lib
+    try {
+      const tokens = await readOnlyWalletService.getTokens();
+      const tokenData = tokens.find((t: Record<string, unknown>) => t.uid === tokenUid) as Record<string, unknown> | undefined;
```
r4mmer: Is iterating over all tokens the only way to get token data on the wallet-service facade?

packages/web-wallet/src/services/TokenStorageService.ts
```diff
@@ -0,0 +1,164 @@
...
+/**
+ * Service for persisting token registrations to localStorage with network-specific keys.
+ *
+ * Storage format:
+ * Key: `hathor_wallet_registered_tokens_{network}_{genesisHash}`
+ * Value: JSON serialized TokenStorageData
+ *
+ * Note: genesisHash is currently empty string, will be populated when RPC handler is updated
+ */
+export class TokenStorageService {
```
r4mmer: Maybe this service should have an in-memory copy and keep the in-memory and localStorage synced.
in-memory fetches are way faster.

packages/web-wallet/src/utils/addressMode.ts
```diff
@@ -0,0 +1,125 @@
...
+export async function getAddressForMode(
+  mode: AddressMode,
+  readOnlyWalletService: ReadOnlyWalletService
+): Promise<string> {
+  if (mode === 'single') {
+    // Always use address at index 0
+    const addressInfo = await readOnlyWalletService.getAddressAtIndex(0);
+    if (!addressInfo) {
+      throw new Error('Failed to get address at index 0');
+    }
```
r4mmer: How? And why only this path has error management

packages/web-wallet/src/utils/addressMode.ts
```diff
@@ -0,0 +1,125 @@
...
+}
+
+/**
+ * Get address info (address + index) based on the selected address mode
+ *
+ * @param mode - 'single' for always using address at index 0, 'dynamic' for generating new addresses
+ * @param readOnlyWalletService - The wallet service instance
+ * @returns Object with address and index
+ */
+export async function getAddressInfoForMode(
```
r4mmer: Maybe the getAddressForMode can use this method internally instead.

packages/web-wallet/src/components/CreateTokenDialog.tsx
```diff
@@ -0,0 +1,561 @@
...
+    .regex(/^\d+$/, 'Amount must be a whole number')
+    .refine((val) => {
+      const num = parseInt(val, 10);
+      return num > 0;
+    }, 'Amount must be greater than 0')
+    .refine((val) => {
+      const num = parseInt(val, 10);
+      return num <= Number.MAX_SAFE_INTEGER;
+    }, 'Amount is too large'),
+  tokenType: z.enum(['deposit', 'fee']),
```
r4mmer: Oh, we are already accounting for fee based tokens?
We should update all references of the config strings to include the token info version

packages/web-wallet/src/components/HistoryDialog.tsx
```diff
@@ -0,0 +1,410 @@
...
+interface ProcessedTransaction {
+  id: string
+  type: 'sent' | 'received'
+  amount: bigint
+  timestamp: string
+  txHash: string
+  status: 'confirmed' | 'pending'
+}
+
+const HistoryDialog: React.FC<HistoryDialogProps> = ({ isOpen, onClose, tokenUid, onRegisterTokenClick }) => {
```
r4mmer: Maybe we should have a docstring explaining what the dialog is about

r4mmer (changes requested):
Why are there so many "fake tests"? Maybe we should create actual tests or at least remove the test files so we do not commit meaningless files.

We should also add the Hathor header on all files.

r4mmer (changes requested):
Why are there so many "fake tests"? Maybe we should create actual tests or at least remove the test files so we do not commit meaningless files.

We should also add the Hathor header on all files.

packages/web-wallet/src/components/ConnectionLostModal.tsx
```diff
@@ -0,0 +1,55 @@
...
+  isOpen: boolean;
+  onReconnect: () => void;
+}
+
+export const ConnectionLostModal: React.FC<ConnectionLostModalProps> = ({
+  isOpen,
+  onReconnect
+}) => {
+  return (
+    <Dialog open={isOpen} onOpenChange={() => {}}>
```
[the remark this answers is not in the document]
andreabadesso: Removed

packages/web-wallet/playwright.config.ts
andreabadesso: Removed those... This was from an early attempt to make playwright tests

But the current PR is big enough as is, we can go back to this in a later PR

packages/web-wallet/README.md
andreabadesso: Done in 3418382f839a42bcc6abd82664bce1142232f411

Dockerfile.web-wallet
```diff
@@ -0,0 +1,45 @@
+# Build stage
+FROM node:20-alpine AS builder
+
+WORKDIR /app
+
+# Copy yarn configuration and patches
+COPY .yarnrc.yml ./
+COPY .yarn ./.yarn
+COPY package.json yarn.lock ./
```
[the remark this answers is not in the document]
andreabadesso: Done

packages/web-wallet/src/contexts/WalletContext.tsx
andreabadesso: Added documentation on be48e1a7d662b24de2793d1ee84899ef84aff3e2

What do you think?

packages/web-wallet/src/components/ui/dialog.tsx
andreabadesso: This is a stylistic choice from `shadcn` components, I can remove it if you want, but it's inoffensive

packages/web-wallet/src/components/ErrorNotification.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+interface ErrorNotificationProps {
+  error: Error | null;
+  onDismiss: () => void;
+}
+
+/**
+ * Error notification component that displays errors in a toast-like notification.
+ * Auto-dismisses after 10 seconds but can be manually dismissed.
+ */
```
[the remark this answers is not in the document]
andreabadesso: We don't. I removed it and forgot to remove the docs

packages/web-wallet/src/components/Header.tsx
andreabadesso: Yes, it works

packages/web-wallet/src/components/Header.tsx
```diff
@@ -0,0 +1,249 @@
...
+    if (address) {
+      try {
+        await navigator.clipboard.writeText(address);
+        setCopiedAddress(true);
+        setTimeout(() => setCopiedAddress(false), 2000);
+      } catch (error) {
+        console.error('Failed to copy address:', error);
+      }
+    }
+  };
```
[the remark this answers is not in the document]
andreabadesso: Hey, it was indeed glitched, it would "blink" if I pressed multiple times, thanks!

packages/web-wallet/src/components/Header.tsx
```diff
@@ -0,0 +1,249 @@
...
+    setIsDisconnectModalOpen(true);
+  };
+
+  const handleDisconnect = () => {
+    setIsMenuOpen(false);
+    setIsDisconnectModalOpen(false);
+    disconnectWallet();
+    // Page will reload automatically via state reset
+    window.location.reload();
+  };
```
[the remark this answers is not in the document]
andreabadesso: You're right, removed the reload

packages/web-wallet/src/components/RegisterTokenDialog.tsx
```diff
@@ -0,0 +1,203 @@
...
+            {errors.configString && (
+              <div className="flex items-start gap-2 mt-2">
+                <AlertCircle className="w-4 h-4 text-red-400 flex-shrink-0 mt-0.5" />
+                <span className="text-xs text-red-400">{errors.configString.message}</span>
+              </div>
+            )}
+            {/* Only show format hint after user starts typing */}
+            {configString && (
+              <p className="text-xs text-muted-foreground mt-2">
+                Format: [name:symbol:uid:checksum]
```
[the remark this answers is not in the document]
andreabadesso: We'll handle fee based token in the next version

packages/web-wallet/src/components/RegisterTokenDialog.tsx
```diff
@@ -0,0 +1,203 @@
...
+            <div className="flex items-start gap-2 p-3 bg-green-500/10 border border-green-500/50 rounded-lg">
+              <CheckCircle className="w-4 h-4 text-green-400 flex-shrink-0 mt-0.5" />
+              <span className="text-green-400 text-sm">Token registered successfully!</span>
+            </div>
+          )}
+
+          {/* Register Button */}
+          <button
+            type="submit"
+            disabled={isLoading || !configString || !!errors.configString}
```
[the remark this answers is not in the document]
andreabadesso: No, it means that there are errors found, so the submit button will be disabled when `!!errors.configString` returns `true`

packages/web-wallet/src/components/__tests__/CreateTokenDialog.test.tsx
```diff
@@ -0,0 +1,259 @@
+import { describe, it, expect } from 'vitest';
+
+describe('CreateTokenDialog - Financial Validation', () => {
+  describe('Deposit Calculation', () => {
+    it('should calculate 1 cent per token deposit correctly', () => {
+      // Formula: 1 cent per token
+      const calculateDeposit = (tokenAmount: number): bigint => {
+        return BigInt(Math.ceil(tokenAmount));
+      };
```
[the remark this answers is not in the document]
andreabadesso: Useless test deleted, using `tokenUtils`

packages/web-wallet/src/components/ui/button.tsx
andreabadesso: As I said before, this is a stylistic choice from shadcn components

packages/web-wallet/src/contexts/hooks/useNetworkManagement.ts
```diff
@@ -0,0 +1,238 @@
...
+          newNetwork: newNetwork,
+        }
+      });
+
+      onLoadingChange(true, 'Stopping previous wallet...');
+
+      // Stop the old read-only wallet
+      if (readOnlyWalletService.isReady()) {
+        await readOnlyWalletService.stop();
+      }
```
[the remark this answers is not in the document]
andreabadesso: Refactored this so the wallet lifecycle is encapsulated in the `useWalletConnection` hook

packages/web-wallet/src/contexts/WalletContext.tsx
andreabadesso: Fixed, thanks

packages/web-wallet/src/contexts/hooks/useTransactions.ts
```diff
@@ -0,0 +1,97 @@
...
+      log.info('Transaction completed successfully');
+      return result;
+    } catch (error) {
+      log.error('Transaction failed in wrapper:', error);
+      onSnapError(error);
+    }
+  };
+
+  const setHistoryDialogState = (isOpen: boolean, page: number = 0) => {
+    setIsHistoryDialogOpen(isOpen);
```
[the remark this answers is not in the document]
andreabadesso: Refactored

packages/web-wallet/src/contexts/hooks/useWalletConnection.ts
```diff
@@ -0,0 +1,502 @@
...
+        const networkTest = await Promise.race([networkCheckPromise, networkCheckTimeout]);
+
+        let parsedNetworkTest;
+        try {
+          parsedNetworkTest = typeof networkTest === 'string' ? JSON.parse(networkTest) : networkTest;
+        } catch (parseError) {
+          console.error('Failed to parse network test response:', parseError, networkTest);
+          throw new Error(`Snap returned invalid JSON response: ${parseError instanceof Error ? parseError.message : 'Parse error'}`);
+        }
+        currentSnapNetwork = parsedNetworkTest?.response?.network;
```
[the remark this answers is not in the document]
andreabadesso: Yes, much better

Thanks

packages/web-wallet/src/contexts/hooks/useWalletConnection.ts
andreabadesso: Refactored them in 4fc0b319d1fcc2113885d1157887d37e41ed380f

Much better, thanks

packages/web-wallet/src/contexts/hooks/useWalletConnection.ts
```diff
@@ -0,0 +1,502 @@
...
+            params: {
+              network: currentSnapNetwork,
+              newNetwork: targetNetwork
+            }
+          });
+          const changeNetworkTimeout = new Promise((_, reject) =>
+            setTimeout(() => reject(new Error('Network change timeout')), SNAP_TIMEOUTS.NETWORK_CHANGE)
+          );
+          await Promise.race([changeNetworkPromise, changeNetworkTimeout]);
+        } catch (networkError) {
```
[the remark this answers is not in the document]
andreabadesso: Refactored to add an util

packages/web-wallet/src/contexts/hooks/useWalletConnection.ts
```diff
@@ -0,0 +1,502 @@
...
+      const newXpub = (parsedResponse as { response?: { xpub?: string } })?.response?.xpub;
+
+      if (!newXpub) {
+        throw new Error('Failed to get xpub from snap - no xpub in response');
+      }
+
+      setLoadingStep('Initializing read-only wallet...');
+
+      const newNetwork = DEFAULT_NETWORK;
+      await readOnlyWalletService.initialize(newXpub, newNetwork);
```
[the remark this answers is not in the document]
andreabadesso: Explained above

packages/web-wallet/src/contexts/hooks/useWalletConnection.ts
```diff
@@ -0,0 +1,502 @@
...
+      // Verify snap is properly installed and enabled
+      log.debug('Starting snap verification...');
+      setLoadingStep('Verifying snap installation...');
+
+      try {
+        log.debug('Checking installed snaps via wallet_getSnaps...');
+
+        const snaps = await (window as { ethereum?: { request: (args: { method: string }) => Promise<unknown> } }).ethereum?.request({
+          method: 'wallet_getSnaps',
+        }) as Record<string, { version: string; enabled: boolean; blocked: boolean }> | undefined;
```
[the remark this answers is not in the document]
andreabadesso: Ops, we should use the provider from snap-utils indeed. Fixed

packages/web-wallet/src/hooks/useTokens.ts
andreabadesso: image


We use it to display how many custom tokens we have in the Home screen

packages/web-wallet/src/contexts/hooks/useWalletConnection.ts
```diff
@@ -178,191 +192,235 @@ export function useWalletConnection(options: UseWalletConnectionOptions): Wallet
...
-    if (isCheckingRef.current) {
-      return;
+  /**
+   * Verifies snap installation and permissions.
+   * Throws typed errors for different snap states.
+   */
+  const verifySnapInstallation = async (): Promise<void> => {
+    // wallet_getSnaps is a MetaMask wallet method, not a snap method
+    // So we need to call it via window.ethereum directly
+    const snapsResponse = await window.ethereum.request({
```
andreabadesso: This RPC calls metamask directly, not the snap, that's why I'm using `window.ethereum`

packages/web-wallet/src/services/ReadOnlyWalletService.ts
andreabadesso: Removed those

packages/web-wallet/src/services/ReadOnlyWalletService.ts
andreabadesso: Removed

packages/web-wallet/src/services/TokenRegistryService.ts
```diff
@@ -0,0 +1,293 @@
...
+import { readOnlyWalletService } from './ReadOnlyWalletService';
+import { tokenStorageService } from './TokenStorageService';
+import helpers from '@hathor/wallet-lib/lib/utils/helpers';
+import { nftDetectionService } from './NftDetectionService';
+
+/**
+ * Service for managing token registration, validation, and metadata.
+ *
+ * Configuration string format (from wallet-lib):
+ * `[name:symbol:uid:checksum]`
```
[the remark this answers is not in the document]
andreabadesso: Not in this PR

packages/web-wallet/src/services/TokenRegistryService.ts
```diff
@@ -0,0 +1,293 @@
...
+    return tokenInfo;
+  }
+
+  /**
+   * Unregister a token
+   */
+  unregisterToken(tokenUid: string, network: string, genesisHash: string): void {
+    const tokens = tokenStorageService.loadTokens(network, genesisHash);
+    const filtered = tokens.filter(t => t.uid !== tokenUid);
+    const saved = tokenStorageService.saveTokens(network, genesisHash, filtered);
```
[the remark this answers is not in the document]
andreabadesso: Refactored it to be a Map instead of an array

packages/web-wallet/src/services/TokenRegistryService.ts
andreabadesso: This has been heavily refactored, can you check this file again?

packages/web-wallet/src/services/TokenStorageService.ts
andreabadesso: It could, but we already have a in-memory copy in the state so read operations are already very fast:

```
  localStorage (TokenStorageService)
         ↓ load once at startup
  React state (registeredTokens)
         ↓ used for all reads
  UI components / refreshBalance / etc
```

packages/web-wallet/src/utils/addressMode.ts
```diff
@@ -0,0 +1,125 @@
...
+  if (mode === 'single') {
+    // Always use address at index 0
+    const addressInfo = await readOnlyWalletService.getAddressAtIndex(0);
+    if (!addressInfo) {
+      throw new Error('Failed to get address at index 0');
+    }
+    return addressInfo.address;
+  } else {
+    // Dynamic mode: get next available address
+    const addressInfo = readOnlyWalletService.getNextAddress();
```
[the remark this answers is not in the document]
andreabadesso: You are correct, refactored

packages/web-wallet/src/components/CreateTokenDialog.tsx
```diff
@@ -0,0 +1,561 @@
...
+  isNFT: z.boolean(),
+  nftData: z.string().max(150, 'NFT data must be 150 characters or less').optional(),
+  createMintAuthority: z.boolean().optional(),
+  createMeltAuthority: z.boolean().optional(),
```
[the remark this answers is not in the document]
andreabadesso: It already has a default value internally in the hathor-rpc-lib

packages/web-wallet/src/components/AddressModeDialog.tsx
```diff
@@ -0,0 +1,122 @@
...
+                  className="sr-only peer"
+                />
+                <div className="w-5 h-5 border-2 border-border rounded-full bg-[#0D1117] peer-checked:border-primary peer-checked:bg-[#0D1117] transition-all flex items-center justify-center">
+                  <div className="w-2.5 h-2.5 rounded-full bg-primary opacity-0 peer-checked:opacity-100 transition-opacity"></div>
+                </div>
+              </div>
+              <div className="flex-1">
+                <span className="text-sm text-white font-medium block">Single Address</span>
+                <span className="text-xs text-muted-foreground block mt-1">
+                  Always use the same address (index 0)
```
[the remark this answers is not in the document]
andreabadesso: I like it, but not the scope of this PR and is not on figma

packages/web-wallet/src/components/WalletHome.tsx
```diff
@@ -46,6 +44,62 @@ const WalletHome: React.FC = () => {
...
+      return; // This is for history dialog, not notification
+    }
+
+    const notification = tx as { type: 'sent' | 'received'; amount: bigint; timestamp: number };
+    const isReceived = notification.type === 'received';
+
+    toast({
+      variant: isReceived ? 'success' : 'info',
+      title: `${isReceived ? 'Received' : 'Sent'} HTR`,
+      description: `${isReceived ? '+' : '-'}${formatHTRAmount(notification.amount, false)} HTR`,
```
r4mmer: These notifications are only for HTR or can other custom tokens be notified here?

packages/web-wallet/src/hooks/use-toast.ts
```diff
@@ -0,0 +1,193 @@
+"use client"
```
r4mmer: Why are we using this directive, isn't this a static SPA, meaning it's all client side?

packages/web-wallet/src/components/TransactionNotification.tsx
```diff
@@ -0,0 +1,83 @@
...
+  } | null;
+  onDismiss: () => void;
+  onViewHistory: () => void;
+}
+
+/**
+ * Transaction notification component that displays new transactions in a toast-like notification.
+ * Auto-dismisses after 5 seconds but can be manually dismissed.
+ */
+const TransactionNotification: React.FC<TransactionNotificationProps> = ({
```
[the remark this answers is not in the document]
r4mmer: So technically i could time it so the user misses the first transaction?

packages/web-wallet/src/constants.ts
r4mmer: I'm sorry, at some point I saw that testnet explorer was still connected to testnet-golf and not india, this has since been fixed.

packages/web-wallet/src/contexts/hooks/useWalletConnection.ts
```diff
@@ -0,0 +1,502 @@
...
+  }, [isConnected, onRefreshBalance]);
+
+  // Set the ref
+  handleNewTransactionRef.current = handleNewTransaction;
+
+  // Check existing connection on mount
+  useEffect(() => {
+    let timeoutId: NodeJS.Timeout;
+
+    const timer = setTimeout(() => {
```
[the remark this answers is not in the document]
r4mmer: 😎 nice

packages/snap/src/constants.ts
```diff
@@ -52,6 +52,6 @@ export const NETWORK_MAP = {
   'dev-testnet': DEV_TESTNET_URLS,
 }
 
-export const DEFAULT_NETWORK = 'mainnet';
+export const DEFAULT_NETWORK = 'testnet';
```
pedroferreira1: We may forget to change this. Maybe you could change to mainnet in the PR and keep it as testnet locally only

packages/web-wallet/src/utils/hathor.ts
```diff
@@ -0,0 +1,118 @@
...
+};
+
+/**
+ * Truncate address for display
+ * @param address Full address
+ * @param prefixLength Length of prefix to show
+ * @param suffixLength Length of suffix to show
+ * @returns Truncated address
+ */
+export const truncateAddress = (address: string, prefixLength = 7, suffixLength = 7): string => {
```
pedroferreira1: You can use this to truncate also uids. Maybe we could rename to `truncateString` and use in other places

packages/web-wallet/src/utils/hathor.ts
```diff
@@ -0,0 +1,118 @@
...
+ *
+ * @param amount Amount as string (e.g., "10.50")
+ * @returns Amount in cents as BigInt
+ * @throws Error if amount format is invalid
+ */
+export const htrToCents = (amount: string): bigint => {
+  // Validate format - must be digits with optional decimal (max 2 places)
+  const trimmed = amount.trim();
+  if (!/^\d+(\.\d{1,2})?$/.test(trimmed)) {
+    throw new Error('Invalid amount format. Use up to 2 decimal places.');
```
pedroferreira1: Why don't we limit to the `DECIMAL_PLACES` constant?

packages/web-wallet/src/utils/hathor.ts
```diff
@@ -0,0 +1,118 @@
...
+ *
+ * @example
+ * centsToHTR(12345n) // "123.45"
+ * centsToHTR(1000000n) // "10000.00"
+ * centsToHTR(1n) // "0.01"
+ *
+ * @param cents Amount in cents (can be number or BigInt)
+ * @returns Amount as string with decimal point
+ */
+export const centsToHTR = (cents: number | bigint): string => {
```
pedroferreira1: The same as in the other

packages/web-wallet/src/utils/tokenLoading.ts
```diff
@@ -0,0 +1,157 @@
...
+  });
+
+  // Fetch balances for all registered tokens
+  const failedTokens: Array<{ uid: string; symbol: string; error: string }> = [];
+  const tokensWithBalances = await Promise.all(
+    registeredTokens.map(async (token) => {
+      try {
+        const tokenBalances = await readOnlyWalletWrapper.getBalance(token.uid);
+        const balance = tokenBalances.get(token.uid);
+
```
pedroferreira1: We could use the method below `fetchTokenBalance `

packages/web-wallet/src/contexts/hooks/useTokenManagement.ts
```diff
@@ -0,0 +1,134 @@
...
+  const [registeredTokens, setRegisteredTokens] = useState<TokenInfo[]>([]);
+  const [selectedTokenFilter, setSelectedTokenFilter] = useState<TokenFilter>('tokens');
+
+  const refreshTokenBalances = useCallback(async () => {
+    if (!isConnected || !readOnlyWalletWrapper.isReady()) return;
+
+    try {
+      // Fetch balances for all tokens
+      const balanceUpdates = await Promise.all(
+        registeredTokens.map(async (token) => {
```
pedroferreira1: Are we fetching balance for HTR in this case? Or just the other registered tokens?

packages/web-wallet/src/components/CreateTokenDialog.tsx
```diff
@@ -0,0 +1,571 @@
...
+const createTokenSchema = z.object({
+  name: z
+    .string()
+    .min(1, 'Token name is required')
+    .max(30, 'Token name must be 30 characters or less'),
+  symbol: z
+    .string()
+    .min(2, 'Symbol must be at least 2 characters')
+    .max(5, 'Symbol must be 5 characters or less')
+    .regex(/^[A-Z0-9]+$/, 'Symbol must contain only uppercase letters and numbers')
```
pedroferreira1: We have tokens created on mainnet that have lowercase and emoji

packages/web-wallet/src/components/CreateTokenDialog.tsx
```diff
@@ -0,0 +1,571 @@
...
+                          If you want to be able to melt units of this NFT.
+                        </span>
+                      </div>
+                    </label>
+                  </div>
+                </div>
+              )}
+
+              {/* Token Type Dropdown - Only shown for regular tokens */}
+              {!isNFT && (
```
pedroferreira1: We should hide this section and default to deposit

packages/web-wallet/src/components/HistoryDialog.tsx
```diff
@@ -0,0 +1,396 @@
...
+        timestamp: new Date((transaction.timestamp as number) * 1000).toISOString(),
+        txHash: transaction.tx_id as string,
+        status: !(transaction.is_voided as boolean) ? 'confirmed' : 'pending'
+      };
+
+      // Use functional update to avoid race conditions with duplicate check
+      setTransactions(prev => {
+        const isDuplicate = prev.some(tx => tx.id === transaction.tx_id);
+        if (isDuplicate) return prev;
+        return [processedTx, ...prev].slice(0, PAGE_SIZE);
```
pedroferreira1: What if you have already loaded more pages? You go back to load only page 1?

packages/web-wallet/src/components/HistoryDialog.tsx
```diff
@@ -0,0 +1,396 @@
...
+        const amount = balanceValue >= 0n ? balanceValue : -balanceValue
+
+        return {
+          id: tx.tx_id,
+          type,
+          amount,
+          timestamp: new Date(tx.timestamp * 1000).toISOString(),
+          txHash: tx.tx_id,
+          status: !tx.is_voided ? 'confirmed' : 'pending'
+        }
```
pedroferreira1: This is the same code as in the `useEffect` above. Can we reuse it somehow?

packages/web-wallet/src/components/SendDialog.tsx
```diff
@@ -0,0 +1,631 @@
...
+                  />
+                </div>
+              </div>
+            )}
+          </div>
+
+          {/* Send Button */}
+          <button
+            type="submit"
+            disabled={isLoading || !amount}
```
pedroferreira1:
```suggestion
            disabled={isLoading || !amount || !address}
```

packages/web-wallet/src/services/SnapService.ts
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    // Re-throw the original error
+    throw error;
+  }
+}
+
+// These will be used by the wallet context with the hooks
+export const WalletServiceMethods = {
+  async getAddress(invokeSnap: InvokeSnapFunction, type: 'index' = 'index', index: number = 0): Promise<string> {
```
pedroferreira1: It won't throw any errors but the type might be `first_empty`, which does not expect index parameter

packages/web-wallet/src/services/ReadOnlyWalletWrapper.ts
```diff
@@ -0,0 +1,321 @@
...
+  }
+
+  /**
+   * Stop the wallet and cleanup resources
+   */
+  async stop(): Promise<void> {
+    if (!this.wallet) return;
+
+    try {
+      this.wallet.removeAllListeners();
```
pedroferreira1: The `stop` method of the wallet already calls this

packages/web-wallet/src/utils/hathor.ts
```diff
@@ -0,0 +1,118 @@
...
+ *
+ * @example
+ * formatHTRAmount(12345n) // "123.45"
+ * formatHTRAmount(1000000n) // "10,000.00"
+ *
+ * @param amount Amount in cents (satoshis) - can be number or BigInt
+ * @param isNft Whether this is an NFT (uses 0 decimals) or regular token (uses 2 decimals)
+ * @returns Formatted string with decimal places and thousand separators
+ */
+export const formatHTRAmount = (amount: number | bigint, isNft: boolean = false): string => {
```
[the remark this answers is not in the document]
andreabadesso: Refactored

packages/web-wallet/src/contexts/hooks/useTokenManagement.ts
```diff
@@ -0,0 +1,134 @@
...
+      throw error;
+    }
+  };
+
+  const unregisterToken = async (tokenUid: string) => {
+    if (!isConnected) {
+      throw new Error('Wallet is not connected');
+    }
+
+    // Prevent HTR from being unregistered
```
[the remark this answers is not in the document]
andreabadesso: Well, not anymore, added a guard

packages/web-wallet/src/components/WalletHome.tsx
```diff
@@ -0,0 +1,304 @@
...
+                  {customTokenCount > 0 && nftCount > 0 && ' and '}
+                  {nftCount > 0 && `${nftCount} NFT${nftCount > 1 ? 's' : ''}`}
+                </p>
+              )}
+            </div>
+
+            {/* Right: Action Buttons */}
+            <div className='flex gap-3 w-full md:w-auto'>
+              <button
+                onClick={() => navigate('?dialog=send')}
```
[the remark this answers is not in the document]
andreabadesso: What do you mean? Those are different buttons

packages/web-wallet/src/components/WalletHome.tsx
```diff
@@ -0,0 +1,304 @@
...
+              <p className='text-xs text-primary-400 uppercase tracking-wider'>assets summary</p>
+              <div className='flex items-center gap-3'>
+                {/* HTR Icon */}
+                <img
+                  src={htrLogoWhiteOutline}
+                  alt='HTR'
+                  className='w-5 h-5 md:w-6 md:h-6'
+                />
+                <span className='text-xl md:text-2xl font-medium text-white'>
+                  {formatHTRAmount(balances.get('00')?.available ?? 0n, false)} HTR
```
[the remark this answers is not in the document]
andreabadesso: Done

packages/web-wallet/src/components/CreateTokenDialog.tsx
```diff
@@ -0,0 +1,571 @@
...
+
+  const isNFT = watch('isNFT');
+  const amount = watch('amount');
+  const tokenType = watch('tokenType');
+
+  // Calculate 1% HTR deposit (1 HTR per 100 tokens)
+  // For every 100 tokens, 1 HTR (100 cents) deposit is required
+  // Formula: (tokens / 100) * 100 cents = tokens * 1 cent
+  const depositInCents = useMemo(() => {
+    const amountNum = parseInt(amount || '0', 10);
```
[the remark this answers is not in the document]
andreabadesso: Ops, we couldn't, now we can. Fixed

packages/web-wallet/src/components/CreateTokenDialog.tsx
andreabadesso: Done

packages/web-wallet/src/services/SnapService.ts
```diff
@@ -0,0 +1,193 @@
...
+  type?: string;
+  data?: string;
+}
+
+export interface SendTransactionParams {
+  network: string;
+  outputs: TransactionOutput[];
+  inputs?: Array<{ txId: string; index: number }>;
+  changeAddress?: string;
+}
```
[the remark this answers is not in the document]
andreabadesso: Added @hathor/hathor-rpc-handler as a workspace dependency and now usin SendTransactionRpcRequest['params'] from there.

packages/web-wallet/src/services/SnapService.ts
andreabadesso: Updated to handle first_empty type, only includes index param when type === 'index'

packages/web-wallet/src/services/RegisteredTokenStorageService.ts
```diff
@@ -0,0 +1,248 @@
...
+      const key = this.getDataStorageKey(network, genesisHash);
+      const stored = localStorage.getItem(key);
+
+      if (!stored) {
+        return {};
+      }
+
+      const data = JSON.parse(stored);
+
+      // Handle migration from old array format
```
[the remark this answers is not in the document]
andreabadesso: We don't. Removed

packages/web-wallet/src/services/ReadOnlyWalletWrapper.ts
andreabadesso: Okay, removed

packages/web-wallet/src/components/TransactionNotification.tsx
```diff
@@ -0,0 +1,83 @@
...
+          <div className={`p-1.5 rounded-lg ${isReceived ? 'bg-green-500/20' : 'bg-blue-500/20'} flex-shrink-0`}>
+            {isReceived ? (
+              <ArrowDownLeft className={`w-4 h-4 ${iconColor}`} />
+            ) : (
+              <ArrowUpRight className={`w-4 h-4 ${iconColor}`} />
+            )}
+          </div>
+          <div className="flex-1 min-w-0">
+            <h3 className={`text-sm font-semibold ${textColor} mb-1`}>
+              {isReceived ? 'Received' : 'Sent'} HTR
```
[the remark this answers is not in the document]
andreabadesso: I commented out this TransactionNotification feature because `checkAddressMine` is currently broken with the read only token

packages/web-wallet/src/components/WalletHome.tsx
andreabadesso: I commented out these notifications because `checkAddressMine` is currently broken

packages/web-wallet/playwright.config.ts
andreabadesso: Issue created: https://github.com/HathorNetwork/hathor-rpc-lib/issues/108

packages/web-wallet/src/components/TransactionNotification.tsx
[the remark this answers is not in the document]
andreabadesso: He already have the transaction history, I don't think we need another history in the app

packages/web-wallet/package.json
```diff
@@ -0,0 +1,71 @@
...
+    "eslint-plugin-react-hooks": "5.2.0",
+    "eslint-plugin-react-refresh": "0.4.20",
+    "events": "3.3.0",
+    "globals": "16.3.0",
+    "jsdom": "27.0.1",
+    "postcss": "8.5.6",
+    "process": "0.11.10",
+    "stream-browserify": "3.0.0",
+    "tailwindcss": "3.4.17",
+    "typescript": "~5.8.3",
```
pedroferreira1: This is the only one not fixed

packages/web-wallet/public/favicon-32x32.png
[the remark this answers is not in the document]
pedroferreira1: Yes but why are they different?

image

packages/web-wallet/src/services/TokenRegistryService.ts
```diff
@@ -0,0 +1,303 @@
...
+        registeredAt: Date.now(),
+      };
+
+      return {
+        ...data,
+        ...metadata,
+        balance: {
+          available: 0n,
+          locked: 0n,
+        },
```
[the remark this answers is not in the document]
pedroferreira1: It's ok, we can keep it

packages/web-wallet/src/components/CreateTokenDialog.tsx
```diff
@@ -127,11 +187,16 @@ const CreateTokenDialog: React.FC<CreateTokenDialogProps> = ({ isOpen, onClose }
       const changeAddress = await getAddressForMode(addressMode, readOnlyWalletWrapper);
       const mintAddress = await getAddressForMode(addressMode, readOnlyWalletWrapper);
 
+      // Convert amount to base units for the RPC
+      const amountInBaseUnits = data.isNFT
+        ? BigInt(data.amount) * BigInt(HTR_DECIMAL_MULTIPLIER)
```
pedroferreira1: In this case we would be creating 100 of the NFT amount? We should create 1

packages/web-wallet/src/components/WalletHome.tsx
```diff
@@ -0,0 +1,312 @@
...
+    );
+  }
+
+  return (
+    <div className="min-h-screen bg-[#0d1117] text-white">
+      <Header
+        onRegisterTokenClick={() => navigate('?dialog=register')}
+        onCreateTokenClick={() => navigate('?dialog=create')}
+        onAddressModeClick={() => navigate('?dialog=addressMode')}
+      />
```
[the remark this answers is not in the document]
andreabadesso: You are right, fixed

packages/web-wallet/src/components/WalletHome.tsx
andreabadesso: I commented out transaction notification for now

packages/web-wallet/package.json
andreabadesso: DOne

packages/snap/snap.manifest.json
```diff
@@ -13,6 +13,10 @@
       }
     }
   },
+  "initialConnections": {
+    "http://localhost:5173": {},
+    "http://localhost:8173": {}
+  },
```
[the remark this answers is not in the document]
andreabadesso: Removed

PR #162, opened 2026-03-31
feat: release utxos after user rejection or error

raul-oliveira:
Requires https://github.com/HathorNetwork/hathor-wallet-lib/pull/1050

### Motivation

The nano contract methods in rpc-lib causes utxos to be indefinitely locked when user rejects a prompt or an error is thrown. 

### Acceptance Criteria
- call release utxos after a prompt rejection
- call release utxos when an error is thrown

### Checklist
- [ ] If you are requesting a merge into `master`, confirm this code is production-ready and can be included in future releases as soon as it gets merged
- [ ] Make sure either the unit tests and/or the QA tests are capable of testing the new features
- [ ] Make sure you do not include new dependencies in the project unless strictly necessary and do not include dev-dependencies as production ones. More dependencies increase the possibility of one of them being hijacked and affecting us.

## Summary by CodeRabbit

* **Bug Fixes**
  * Reserved transaction outputs are now reliably released on user cancellation (confirmation or PIN rejection) and on execution failures, preventing funds from remaining locked for nano-contract and standard transactions.

* **Tests**
  * Added tests asserting cleanup is invoked for cancellation and failure paths across nano-contract creation, nano-contract sends, and standard sends.

### DIFF
--- a/packages/hathor-rpc-handler/__tests__/rpcMethods/sendTransaction.test.ts
+++ b/packages/hathor-rpc-handler/__tests__/rpcMethods/sendTransaction.test.ts
@@ -81,6 +81,7 @@ describe('sendTransaction', () => {
         prepareTx: jest.fn().mockResolvedValue(mockTransaction),
         signTx: jest.fn().mockResolvedValue(mockTransaction),
         runFromMining: sendTransactionMock,
+        releaseUtxos: jest.fn().mockResolvedValue(undefined),
       }),
     } as unknown as jest.Mocked<IHathorWallet>;
 
@@ -417,6 +418,7 @@ describe('sendTransaction', () => {
       prepareTx: jest.fn().mockResolvedValue(mockPreparedTransaction),
       signTx: signTxMock,
       runFromMining: sendTransactionMock,
+      releaseUtxos: jest.fn().mockResolvedValue(undefined),
     });
 
     promptHandler
@@ -464,6 +466,7 @@ describe('sendTransaction', () => {
       prepareTx: jest.fn().mockResolvedValue(mockTransaction),
       signTx: signTxMock,
       runFromMining: sendTransactionMock.mockResolvedValue(txResponse),
+      releaseUtxos: jest.fn().mockResolvedValue(undefined),
     });
 
     promptHandler
@@ -510,6 +513,7 @@ describe('sendTransaction', () => {
       prepareTx: jest.fn().mockResolvedValue(fbtMockTransaction),
       signTx: jest.fn().mockResolvedValue(fbtMockTransaction),
       runFromMining: sendTransactionMock.mockResolvedValue({ hash: 'txHash123' }),
+      releaseUtxos: jest.fn().mockResolvedValue(undefined),
     });
 
     promptHandler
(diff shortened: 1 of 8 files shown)


packages/hathor-rpc-handler/__tests__/rpcMethods/sendTransaction.test.ts
```diff
@@ -539,4 +543,77 @@ describe('sendTransaction', () => {
       {},
     );
   });
+
+  it('should call releaseUtxos when user rejects confirmation prompt', async () => {
```
tuliomir, line 547: suggestion(non-blocking): Is there a way to share the test code between those three test files? They validate almost the same thing, but need to have repeated code between files.

yarn.lock
```diff
@@ -2456,7 +2456,7 @@ __metadata:
   resolution: "@hathor/hathor-rpc-handler@workspace:packages/hathor-rpc-handler"
   dependencies:
     "@eslint/js": "npm:9.4.0"
-    "@hathor/wallet-lib": "npm:2.17.0"
+    "@hathor/wallet-lib": "npm:3.0.1"
```
tuliomir, line 2459: question: I noticed the `yarn.lock` didn't remove code from the wallet lib `2.17.0`, which suggests it is still being used by some other part of this repo.

Should we keep it like this, or enforce the new v3 of the wallet lib to all packages at once?

tuliomir (dismissed):

packages/hathor-rpc-handler/src/rpcMethods/sendTransaction.ts
```diff
@@ -225,6 +228,7 @@ export async function sendTransaction(
       response,
     } as RpcResponse;
   } catch (err) {
+    try { await sendTransactionObject.releaseUtxos(); } catch { /* best-effort */ }
```
pedroferreira1: Why did you do this only here? I believe all places should have a `try ... catch`, otherwise it wouldn't throw `PromptRejectedError` and the UI of the wallets would be broken

raul-oliveira: I checked `releaseUtxos()` in `wallet-lib` and it's already best-effort by contract on both sides, so it never throws:

- **`HathorWallet.releaseUtxos()`** wraps each per-UTXO release in its own `try/catch` and logs at `debug` level.
- **`HathorWalletServiceWallet.releaseUtxos()`** is a no-op (UTXO state is managed server-side).

So `PromptRejectedError` is never masked. The `try/catch` I had in the error path was actually redundant — I removed it so all 9 call sites now follow the same simple `await` pattern.

pedroferreira1 (approved):

yarn.lock
raul-oliveira: We should keep it for this PR. Each workspace pins its own version:

- `@hathor/hathor-rpc-handler` → `3.0.1` (this is where the `releaseUtxos`/prepare-then-sign work lands)
- `@hathor/snap` → `2.17.0`
- `@hathor/web-wallet` → `2.14.0`

So `yarn.lock` legitimately resolves three versions because three workspaces depend on three different majors. The scope of this PR is the rpc-handler only, and bumping the snap and web-wallet to v3 has its own surface area (API changes, regression testing, etc.) that I'd rather handle in a dedicated PR.

tuliomir (approved):

PR #166, opened 2026-05-28
feat(web-wallet): enable fee-based tokens and friendly snap errors

raul-oliveira:
### Motivation

Enable the fee based tokens and validates the creation, and send tokens flows.

### Acceptance Criteria

- Token Type dropdown (Deposit / Fee) is visible on Create Token for non-NFT tokens
- Creating with Type=Fee on the explorer shows `FEE MODEL: Fee-based` (the snap actually receives `version: 'fee'`)
- Selecting Fee hides the `DEPOSIT` row, hides the insufficient-balance warning, and keeps the Create button enabled with 0 HTR
- Deposit math is unchanged: `1.00 token = 0.01 HTR` deposit (1%), proportional for larger amounts
- HTR shortage on a fee-based token submit shows "Insufficient HTR to cover the network fee." (both inline dialog and toast), not the raw wallet-lib message
- Generic shortage for other tokens shows "Insufficient balance to send this transaction."
- Unregistering a token surfaces it in the Import Tokens dialog without requiring a reload, new transaction, or reconnect
- After importing a token, opening Send for it shows the real `Balance available` instead of `0.00`
- HTR balance display and behavior are unchanged

### Screenshots
image
image

### Checklist

## Summary by CodeRabbit

* **New Features**
  * Added fee token creation option and exposed token-type selection in the create dialog.
  * Custom tokens now show live balance updates.

* **Bug Fixes**
  * Improved and contextualized user-facing error messages (native vs custom token shortages) across send, create, and connection flows.
  * Token discovery now refreshes when tokens are registered or unregistered.

* **Tests**
  * Added tests covering snap error classification and message mapping.

### DIFF
--- a/packages/web-wallet/src/components/CreateTokenDialog.tsx
+++ b/packages/web-wallet/src/components/CreateTokenDialog.tsx
@@ -1,5 +1,5 @@
 import React, { useState, useMemo, useRef } from 'react';
-import { X, Loader2, AlertCircle, CheckCircle, Copy } from 'lucide-react';
+import { X, Loader2, AlertCircle, CheckCircle, Copy, Info } from 'lucide-react';
 import { useForm } from 'react-hook-form';
 import { zodResolver } from '@hookform/resolvers/zod';
 import { z } from 'zod';
@@ -9,9 +9,9 @@ import { helpersUtils, tokensUtils, constants } from '@hathor/wallet-lib';
 import { formatAmount, amountToCents } from '../utils/hathor';
 import { readOnlyWalletWrapper } from '../services/ReadOnlyWalletWrapper';
 import { getAddressForMode } from '../utils/addressMode';
-// TODO: Re-enable when fee token feature is ready
-// import { Select, SelectContent, SelectItem, SelectTrigger, SelectValue } from './ui/select';
+import { Select, SelectContent, SelectItem, SelectTrigger, SelectValue } from './ui/select';
 import { useToast } from '@/hooks/use-toast';
+import { extractErrorMessage, getSnapErrorUserMessage } from '../utils/snapErrors';
 
 interface CreateTokenDialogProps {
   isOpen: boolean;
@@ -117,7 +117,6 @@ const CreateTokenDialog: React.FC<CreateTokenDialogProps> = ({ isOpen, onClose }
     watch,
     setValue,
     getValues,
-    // TODO: Re-enable when fee token feature is ready
     reset,
   } = useForm<CreateTokenFormData>({
     resolver: zodResolver(createTokenSchema),
@@ -137,16 +136,18 @@ const CreateTokenDialog: React.FC<CreateTokenDialogProps> = ({ isOpen, onClose }
   const isNFT = watch('isNFT');
   const amount = watch('amount');
   const nftData = watch('nftData');
-  // TODO: Re-enable when fee token feature is ready
-  // const tokenType = watch('tokenType');
+  const tokenType = watch('tokenType');
 
-  // When toggling to NFT mode, strip any decimal portion from the amount
+  // When toggling to NFT mode, strip any decimal portion from the amount and
+  // snap the token type to 'deposit' (NFTs are always deposit-based; the
+  // submit handler already forces this, but the UI should reflect it too).
   React.useEffect(() => {
     if (isNFT) {
       const currentAmount = getValues('amount');
       if (currentAmount.includes('.') || currentAmount.includes(',')) {
         setValue('amount', currentAmount.split(/[.,]/)[0]);
       }
+      setValue('tokenType', 'deposit');
     }
   // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [isNFT]);
@@ -158,6 +159,9 @@ const CreateTokenDialog: React.FC<CreateTokenDialogProps> = ({ isOpen, onClose }
   const depositInCents = useMemo(() => {
     try {
       if (!amount || amount === '0') return 0n;
+
+      // Fee tokens don't require an HTR deposit
+      if (!isNFT && tokenType === 'fee') return 0n;
 
       let amountInBaseUnits: bigint;
       if (isNFT) {
@@ -182,7 +186,7 @@ const CreateTokenDialog: React.FC<CreateTokenDialogProps> = ({ isOpen, onClose }
     } catch {
       return 0n;
     }
-  }, [amount, isNFT, nftData]);
+  }, [amount, isNFT, nftData, tokenType]);
 
   // Check if user has insufficient balance
   const hasInsufficientBalance = useMemo(() => {
@@ -210,11 +214,15 @@ const CreateTokenDialog: React.FC<CreateTokenDialogProps> = ({ isOpen, onClose }
         ? BigInt(data.amount)
         : amountToCents(data.amount);
 
+      // NFTs are always deposit-based; only regular tokens can be fee-based
+      const tokenVersion = data.isNFT ? 'deposit' : data.tokenType;
+
       // Prepare RPC params matching createTokenRpcSchema
       const params = {
         name: data.name,
         symbol: data.symbol,
         amount: String(amountInBaseUnits),
+        version: tokenVersion,
         change_address: changeAddress,
         create_mint: createMint,
         create_melt: createMelt,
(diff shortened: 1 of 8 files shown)


packages/web-wallet/src/components/CreateTokenDialog.tsx
```diff
@@ -596,17 +609,9 @@ const CreateTokenDialog: React.FC<CreateTokenDialogProps> = ({ isOpen, onClose }
                       <SelectItem value="fee">Fee</SelectItem>
                     </SelectContent>
                   </Select>
-                  <ul className="list-disc list-inside text-xs text-muted-foreground mt-2 space-y-1">
```
raul-oliveira, line 599: I removed these lines because the tooltip already covers this info.

packages/web-wallet/src/utils/snapErrors.ts
```diff
@@ -102,6 +106,48 @@ export function isSnapDisabledError(errorMessage: string): boolean {
   );
 }
 
+/**
```
raul-oliveira, line 109: This file holds some conditions need to map the error to user friendly messages from wallet-lib. Let me know if we should forward the wallet-lib errors directly instead of wrapping it to the users.

packages/web-wallet/src/utils/snapErrors.ts
```diff
@@ -124,5 +170,16 @@ export function getSnapErrorUserMessage(errorMessage: string): string {
     return 'Snap is disabled. Please enable it in MetaMask settings.';
   }
 
+  // wallet-lib raises "No UTXOs available for the token <uid>." when
+  // auto-selection can't cover the amount. HTR shortage gets its own message
+  // because, on fee-based token operations, it means insufficient HTR for the
+  // network fee.
+  if (isHtrUtxoShortage(errorMessage)) {
+    return 'Insufficient HTR to cover the network fee.';
+  }
```
pedroferreira1, line 206: If the user is sending a tx of HTR and we don't have balance. Will it reach this condition here?

If yes, we shouldn't show "to cover the network fee." It's the next error message

packages/web-wallet/src/hooks/useTokens.ts
```diff
@@ -33,10 +33,19 @@ export function useTokens(selectedTokenFilter: TokenFilter = 'all') {
...
-    return [htrToken, ...registeredTokens];
-  }, [htrToken, registeredTokens]);
+    const merged = registeredTokens.map(token => {
+      const current = balances.get(token.uid);
+      return current
+        ? { ...token, balance: { available: current.available, locked: current.locked } }
+        : token;
+    });
+    return [htrToken, ...merged];
+  }, [htrToken, registeredTokens, balances]);
```
pedroferreira1, line 48: Should we do this also for htrToken?

packages/web-wallet/src/components/CreateTokenDialog.tsx
```diff
@@ -279,15 +287,21 @@ const CreateTokenDialog: React.FC<CreateTokenDialogProps> = ({ isOpen, onClose }
...
         setError('Transaction was cancelled. Please try again.');
-      } else if (errorMsg.includes('insufficient') || errorMsg.includes('Insufficient')) {
-        setError('Insufficient HTR balance for deposit and fees.');
       } else if (errorMsg.includes('timeout') || errorMsg.includes('timed out')) {
         setError('Request timed out. Please check your connection and try again.');
+      } else if (mappedMsg !== errorMsg) {
+        // getSnapErrorUserMessage produced a friendly message — use it.
+        setError(mappedMsg);
+      } else if (errorMsg.includes('insufficient') || errorMsg.includes('Insufficient')) {
+        setError('Insufficient HTR balance for deposit and fees.');
```
pedroferreira1: Isn't it considered in the utils method you created? `isHtrUtxoShortage`?

packages/web-wallet/src/utils/snapErrors.ts
raul-oliveira: Good catch — you're right. `getSnapErrorUserMessage` is reused by `SendDialog`, so an HTR send hitting a UTXO shortage (e.g. balance changing between validation and submit) would surface "Insufficient HTR to cover the network fee.", which is misleading.

The helper now takes an optional `tokenVersion` (reusing the `TokenVersion` enum from `@hathor/wallet-lib`) and routes the HTR-shortage branch accordingly:

- `TokenVersion.NATIVE` → "Insufficient balance to send this transaction."
- `TokenVersion.DEPOSIT` → "Insufficient HTR balance for deposit."
- `TokenVersion.FEE` or no context → keeps the existing network-fee message.

`SendDialog` and `CreateTokenDialog` were updated to pass the right version. Fixed in https://github.com/HathorNetwork/hathor-rpc-lib/pull/166/commits/4ecaf35ba839af09de471506e8464b241c1d92cb.

packages/web-wallet/src/hooks/useTokens.ts
raul-oliveira: Already handled — `htrToken` is built directly from the `balances` Map (lines 21–34), and the `useMemo` re-runs whenever `balances` changes, so it always carries the live value.

The merge added here is only needed for `registeredTokens` (custom tokens), which carry the balance frozen at registration time — those are the ones that needed the overlay.

packages/web-wallet/src/components/CreateTokenDialog.tsx
raul-oliveira: Good point — and it actually exposed a related bug: for deposit-token creation, the mapped message ("…to cover the network fee.") was also misleading since the shortage is the 1% deposit.

`getSnapErrorUserMessage` now takes a `TokenVersion`:

- `TokenVersion.DEPOSIT` → "Insufficient HTR balance for deposit."
- `TokenVersion.FEE` → keeps "Insufficient HTR to cover the network fee." (correct for fee tokens).

With that the `insufficient/Insufficient` fallback became dead code for the UTXO error and was removed. Fixed in https://github.com/HathorNetwork/hathor-rpc-lib/pull/166/commits/4ecaf35ba839af09de471506e8464b241c1d92cb.

pedroferreira1 (approved):

packages/web-wallet/src/components/CreateTokenDialog.tsx
```diff
@@ -137,16 +136,18 @@ const CreateTokenDialog: React.FC<CreateTokenDialogProps> = ({ isOpen, onClose }
...
+  // When toggling to NFT mode, strip any decimal portion from the amount and
+  // snap the token type to 'deposit' (NFTs are always deposit-based; the
+  // submit handler already forces this, but the UI should reflect it too).
   React.useEffect(() => {
     if (isNFT) {
       const currentAmount = getValues('amount');
       if (currentAmount.includes('.') || currentAmount.includes(',')) {
         setValue('amount', currentAmount.split(/[.,]/)[0]);
       }
+      setValue('tokenType', 'deposit');
```
tuliomir, line 150: suggestion(non-blocking): Use constants or an `enum` instead of hardcoded `'deposit'` and `'fee'` instances

packages/web-wallet/src/utils/snapErrors.ts
```diff
@@ -102,13 +106,75 @@ export function isSnapDisabledError(errorMessage: string): boolean {
...
+ * Handles Error instances, JSON-RPC-style `{message: string}` objects, and
+ * anything else by falling back to the provided default.
+ */
+export function extractErrorMessage(err: unknown, fallback: string): string {
+  if (err instanceof Error && err.message) return err.message;
+  if (typeof err === 'object' && err !== null && 'message' in err) {
+    const m = (err as { message: unknown }).message;
+    if (typeof m === 'string' && m.length > 0) return m;
+  }
+  return fallback;
```
tuliomir, line 148: suggestion(non-blocking): Make `fallback` optional and default to `Unknown error` or something similarly generic.

tuliomir (approved):

packages/web-wallet/src/components/CreateTokenDialog.tsx
raul-oliveira: I'll address it after the merge in https://github.com/HathorNetwork/hathor-rpc-lib/issues/170

packages/web-wallet/src/utils/snapErrors.ts
raul-oliveira: The fallback is provided by the caller to have a bit more context accuracy in the message

PR #167, opened 2026-05-29
feat(web-wallet): gate fee-based tokens behind a per-network Unleash flag

raul-oliveira:
### Motivation

The web-wallet should let users create fee-based tokens, but only on **testnet** and behind a feature flag, so the feature can be rolled out gradually. The`fee-based-tokens-webwallet.rollout` Unleash flag is constrained to `network = testnet`, but the web-wallet never told Unleash which network it was on — so the flag never resolved and the feature stayed hidden on every environment. This PR sends the active network to Unleash and gates the fee-based token UI and flow behind the per-network flag.

> Stacked on top of #166 (`raul-oliveira/feat/fee-based-tokens-activation`) — review/merge that one first.

### Acceptance Criteria

- Fee-based token creation is available only when the wallet is connected the `fee-based-tokens-webwallet.rollout` flag is enabled.
- The web-wallet reports its active network to Unleash, so the flag's `network = testnet` constraint is actually evaluated (previously it never matched on any environment).
- On testnet with the flag enabled, the Create Token dialog shows the **Token Type (Deposit / Fee)** selector and lets the user create a fee-based token — no HTR deposit required, sending `version: 'fee'` in the create-token request.
- On **mainnet**, or whenever the flag is disabled, the Token Type selector is hidden and token creation behaves exactly as before the feature (deposit tokens only).
- Switching the network at runtime enables or disables the feature immediately, without reloading the page.
- The web-wallet flag is independent from hathor-wallet-mobile's `fee-based-tokens.rollout`, so each client can be rolled out separately.

### Checklist
- [ ] If you are requesting a merge into `master`, confirm this code is production-ready and can be included in future releases as soon as it gets merged _(N/A — targets the stacked feature branch, not `master`)_

## Summary by CodeRabbit

* **New Features**
  * Introduced fee-based tokens feature toggle (currently disabled by default).
  * Added centralized network management for Hathor network selection with persistent storage.

* **Chores**
  * Updated token creation dialog to conditionally display based on feature toggle status.
  * Refactored provider hierarchy to integrate network context throughout the application.
  * Updated test suite to reflect new provider structure.

### DIFF
--- a/packages/web-wallet/src/App.tsx
+++ b/packages/web-wallet/src/App.tsx
@@ -1,6 +1,7 @@
 import { BrowserRouter, Routes, Route, useNavigate, useParams } from 'react-router-dom'
 import { MetaMaskProvider } from '@hathor/snap-utils'
 import { WalletProvider } from './contexts/WalletContext'
+import { NetworkProvider } from './contexts/NetworkContext'
 import { FeatureToggleProvider, useFeatureToggle } from './contexts/FeatureToggleContext'
 import WalletHome from './components/WalletHome'
 import MaintenancePage from './components/MaintenancePage'
@@ -117,9 +118,11 @@ function App() {
   return (
     <ErrorBoundary>
       <BrowserRouter>
-        <FeatureToggleProvider>
-          <FeatureGatedWallet />
-        </FeatureToggleProvider>
+        <NetworkProvider>
+          <FeatureToggleProvider>
+            <FeatureGatedWallet />
+          </FeatureToggleProvider>
+        </NetworkProvider>
       </BrowserRouter>
     </ErrorBoundary>
   )
--- a/packages/web-wallet/src/components/CreateTokenDialog.tsx
+++ b/packages/web-wallet/src/components/CreateTokenDialog.tsx
@@ -4,6 +4,7 @@ import { useForm } from 'react-hook-form';
 import { zodResolver } from '@hookform/resolvers/zod';
 import { z } from 'zod';
 import { useWallet } from '../contexts/WalletContext';
+import { useFeatureToggle } from '../contexts/FeatureToggleContext';
 import { useInvokeSnap } from '@hathor/snap-utils';
 import { helpersUtils, tokensUtils, constants, TokenVersion } from '@hathor/wallet-lib';
 import { formatAmount, amountToCents } from '../utils/hathor';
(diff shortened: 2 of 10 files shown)


pedroferreira1 (dismissed):

packages/web-wallet/src/components/CreateTokenDialog.tsx
```diff
@@ -138,6 +140,11 @@ const CreateTokenDialog: React.FC<CreateTokenDialogProps> = ({ isOpen, onClose }
   const nftData = watch('nftData');
   const tokenType = watch('tokenType');
 
+  // Fee-based tokens require both the Unleash flag and a non-NFT 'fee' selection.
+  // When the flag is off the Token Type dropdown is hidden, so tokenType stays
+  // 'deposit' and the token behaves as a regular deposit token.
+  const isFeeBasedToken = isFeeTokensEnabled && !isNFT && tokenType === 'fee';
```
tuliomir, line 146: suggestion(non-blocking): Could we avoid hardocoded strings by using constants/enums instead?

packages/web-wallet/src/App.tsx
```diff
@@ -117,9 +118,11 @@ function App() {
   return (
     <ErrorBoundary>
       <BrowserRouter>
-        <FeatureToggleProvider>
-          <FeatureGatedWallet />
-        </FeatureToggleProvider>
+        <NetworkProvider>
+          <FeatureToggleProvider>
+            <FeatureGatedWallet />
```
tuliomir, line 123: thought: I don't think this PR is the correct place to address this issue, but I'd like to raise it at the moment it appeared.

We're having multiple providers for every screen, with information that is necessary to all of them. This is a job `Redux Toolkit` does very well, and it's already a stable pattern in our JS/TS repositories.

I think we should migrate to using Redux Toolkit `slices` instead of having multiple providers on every component and on every test.

tuliomir (dismissed):

packages/web-wallet/src/App.tsx
raul-oliveira: I agree. Also we have zustand that is a modern and simpler way to manage state

raul-oliveira: Issue created https://github.com/HathorNetwork/hathor-rpc-lib/issues/171

packages/web-wallet/src/components/CreateTokenDialog.tsx
raul-oliveira: Issue created in the other pr: https://github.com/HathorNetwork/hathor-rpc-lib/issues/170

tuliomir (approved):

pedroferreira1 (approved):

PR #172, opened 2026-06-17
feat!: prompt-free getBalance for connected dApps with all wallet tokens

raul-oliveira:
### Motivation

Today a dApp must connect and then separately request balance, and **every** `htr_getBalance`
request prompts the user for authorization. This adds UX friction and extra work for dApp
developers. This PR removes the per-request prompt so connected dApps can read balance with a
single frictionless call, and lets them request the balance of **all the tokens the wallet holds**
or just a subset.

Feature requested by Yan — *"Get wallet balance on dApp connection"*.

### Acceptance Criteria

- [x] Balance requests from connected dApps no longer prompt the user (`hathor-rpc-handler` + `snap`)
- [x] dApps can request balance for **all the tokens the wallet holds** (omit `tokens`) **or a subset** (provide `tokens`)
- [x] `htr_getBalance` returns the balance immediately as `GetBalanceObject[]` (response `type` `3`)
- [x] The Snap install dialog discloses balance-read access

### What changed

**`@hathor/hathor-rpc-handler`** (breaking)
- `getBalance` no longer calls `promptHandler` — it returns the balance directly, mirroring the
  existing prompt-free `getConnectedNetwork` / `getWalletInformation` handlers.
- `tokens` is now **optional**; when omitted, the handler returns the balance of **every token the
  wallet holds** via a single `wallet.getBalance(null)` call. When provided, it returns that subset.
- The notion of *registered tokens* belongs to the wallet clients, not the RPC library, so the
  handler no longer enumerates `wallet.storage.getRegisteredTokens()` — it lets the wallet-service
  backend resolve the full token set.
- Removed the now-dead `GetBalanceConfirmationPrompt` / `GetBalanceConfirmationResponse` trigger
  types (and their union/enum entries).
- Tests cover: the no-prompt guarantee, the all-tokens path (a single `getBalance(null)` call),
  subset requests, and param validation.

**`@hathor/snap`**
- Removed the `GetBalanceConfirmationPrompt` branch from the prompt handler and deleted
  `dialogs/balance.tsx`.
- Reworded the install dialog's disclosure to cover the balance of all the wallet's tokens
  (dropped the client-specific *registered tokens* wording).

### Breaking changes

This is a **major** change (`4.x` → `5.0.0`): both the public API surface and the `htr_getBalance`
behavior change in ways that require consumers to adapt.

- **Removed exported types** `GetBalanceConfirmationPrompt` and `GetBalanceConfirmationResponse`,
  plus their members in the `TriggerTypes` / `TriggerResponseTypes` enums and the
  `Trigger` / `TriggerResponse` unions. Any consumer importing these will fail to compile.
- **`getBalance` no longer prompts** for confirmation. Consumers relying on the per-request prompt
  (mobile, desktop) must drop their balance modals.
- `tokens` changed from required to **optional**.

### Notes / rollout

- This is the **rpc-lib half**. The wallet consumers (mobile, desktop) drop their balance modals,
  and the `hathor-rpc-handler` `5.0.0` and `snap` version bumps land, in **separate PRs** (per the
  repo convention of landing version bumps independently from feature changes).
- Omitting `tokens` relies on the wallet facade resolving the full token set: the
  `HathorWalletServiceWallet` facade (used by the snap and the web-wallet) returns every token the
  wallet holds when `getBalance(null)` is called. The direct full-node `HathorWallet` facade does
  not implement the no-token variant, so consumers on that facade should pass an explicit `tokens`
  list.
- The Snap surface is request/response only, so "balance on connect" means the dApp invokes
  `htr_getBalance` right after connecting (now frictionless). Wallet-initiated push
  (`sessionProperties` / `emitSessionEvent`) and a dApp-SDK helper were intentionally **deferred**.

### Checklist

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Balance requests can now omit token IDs to return balances for all registered tokens.
  * Added/extended wallet information to include per-token balances with an optional token filter.
* **Bug Fixes**
  * Balance queries no longer require confirmation prompts.
  * Removed balance-related confirmation dialog and corresponding prompt handling.
* **Documentation**
  * Updated API documentation to reflect optional token parameters and the new wallet information method.
* **Tests / Chores**
  * Refactored RPC test suites for balance and wallet information behavior.
  * Updated Snap manifest checksum and clarified install messaging about all-token balance access.


packages/snap/src/dialogs/install.jsx
```diff
@@ -23,6 +23,9 @@ export const installPage = async () => (
             <Text>
               The dApp you are connecting to will have access to your first address and the network you are connected to
             </Text>
+            <Text>
+              It will also be able to view the balance of your registered tokens.
```
pedroferreira1: We don't have registered tokens for snap. You can use the Text tag above to add this information

"The dApp you are connecting to will have access to your first address, the network you are connected to, and the balance of all your tokens."

packages/hathor-rpc-handler/src/rpcMethods/getBalance.ts
```diff
@@ -61,25 +65,22 @@ export async function getBalance(
...
+  if (!tokenUids) {
+    // getRegisteredTokens() yields only custom tokens, never native HTR — seed the
+    // list with HTR so the "all tokens" response includes it.
+    tokenUids = [constants.NATIVE_TOKEN_UID];
+    for await (const token of wallet.storage.getRegisteredTokens()) {
+      if (token.uid !== constants.NATIVE_TOKEN_UID) {
+        tokenUids.push(token.uid);
+      }
+    }
+  }
```
pedroferreira1: The registered tokens idea is for the wallets (clients). The rpc library shouldn't consider this when implementing its methods. 

We should get all tokens by default, when the user does not send the parameter, but only if the wallet service can handle this request. is it possible for the wallet service facade to return all the tokens the user has in his wallet?

packages/snap/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@hathor/snap",
   "packageManager": "yarn@4.2.2",
-  "version": "0.4.2",
+  "version": "0.5.0",
```
pedroferreira1: We should bump the version in another PR, just like you did for the rpc lib (https://github.com/HathorNetwork/hathor-rpc-lib/pull/173). You can even reuse this for both bumps

packages/snap/src/dialogs/install.jsx
raul-oliveira: Done — collapsed into a single `Text` tag with that wording: "The dApp you are connecting to will have access to your first address, the network you are connected to, and the balance of all your tokens."

packages/hathor-rpc-handler/src/rpcMethods/getBalance.ts
raul-oliveira: Yes, the wallet-service facade can return everything. `getBalance(null)` → `GET /wallet/balances` with no `token_id`, and the backend resolves the full set in one call.

**But `getBalance(null)` is the one call that's wallet-service-only:** the full-node `HathorWallet` throws `WalletError('Not implemented.')` for a null token (`new/wallet.js`). Every other path here passes an explicit token, which both facades support — so only this new no-`tokens` path would break, and only on full-node.

To keep it facade-agnostic, the handler now builds the list from `wallet.getTokens()` (both facades implement it.

packages/snap/package.json
raul-oliveira: Reverted the snap bump (`package.json` + `snap.manifest.json`) back to `0.4.2`; the version bump will land in a separate PR.

pedroferreira1 (dismissed):

packages/hathor-rpc-handler/src/rpcMethods/getBalance.ts
```diff
@@ -61,25 +65,19 @@ export async function getBalance(
...
+    tokens = params.tokens;
+  } else {
+    // getTokens (not getBalance(null), which throws on the fullnode facade) is
+    // the only all-tokens primitive both wallet facades implement.
+    tokens = await wallet.getTokens();
+  }
+
   const balances: GetBalanceObject[] = (await Promise.all(
-    params.tokens.map(token => wallet.getBalance(token)),
+    tokens.map(token => wallet.getBalance(token)),
```
tuliomir: question(non-blocking): You mentioned there is a single API endpoint on the Wallet Service that would return this information already consolidated.

I think it would be better to query this optimized endpoint if it's available: we avoid bumping into eventual rejections from the wallet service in case of a large amount of tokens for the wallet, which would interrupt the loading phase and actually prevent the user from using the wallet at all.

Marking as _non-blocking_ because I think this is a minor improvement: the solid logic is already here - compatible with both facades.

tuliomir (dismissed):

packages/hathor-rpc-handler/src/rpcMethods/getBalance.ts
raul-oliveira: Good idea. I believe we should add a endpoint in the wallet-service to return a batch of tokens. I'll open an issue for it

packages/hathor-rpc-handler/__tests__/rpcMethods/getBalance.test.ts
```diff
@@ -9,185 +9,158 @@ import type { IHathorWallet } from '@hathor/wallet-lib';
...
+      .mockResolvedValueOnce(htrBalance)
+      .mockResolvedValueOnce(customBalance);
+
+    const request = {
       method: RpcMethods.GetBalance,
-      params: {
-        network: 'testnet',
-        tokens: [],
-      },
+      params: { network: 'testnet', tokens: [] },
```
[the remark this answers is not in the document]
raul-oliveira: fixed. I reverted the thing. now if a token array is not provided, the caller validate it against a schema.

pedroferreira1 (approved):

packages/hathor-rpc-handler/src/rpcMethods/getBalance.ts
raul-oliveira: https://github.com/HathorNetwork/hathor-wallet-service/issues/468

tuliomir (approved):

Issue #177, opened 2026-07-06
Web Wallet E2E (MetaMask/Snap): general improvements


tuliomir:
# Web Wallet E2E (MetaMask/Snap): follow-ups after PR #169

Tracking issue for improvements identified while reviewing HathorNetwork/hathor-rpc-lib#169, which added the real-MetaMask-Flask Playwright E2E suite under `packages/web-wallet/tests/e2e/`. **None of these blocked that PR** — it landed intentionally as on-demand, headed, non-gating infrastructure. This issue is the durable home for the next iteration, written against a `main` where #169 has already merged.

Ordered by value. Each item is independent; a future PR can pick off any subset.

---

## 1. Clean checkout can't run the suite from the docs alone — missing dependency build ⛔️ highest value

Following the "How to run" docs literally, from a clean tree, **all four journeys fail at connect.** Root cause: nothing builds the workspace dependency `@hathor/hathor-rpc-handler`. Its `package.json` points `main` at `dist/index.js`, but the repo root `package.json` has no scripts and there is no repo-wide build, so the snap's `mm-snap watch` dev build dies with:

```
Module not found: Can't resolve '@hathor/hathor-rpc-handler'
```

The snap bundle is then broken and every journey fails at `warmSnap` → connect. The missing step:

```bash
yarn workspace @hathor/hathor-rpc-handler build   # produces dist/index.js
```

This matters beyond convenience: the E2E build consumes `@hathor/hathor-rpc-handler` through its built `dist`, so **anyone running the suite without this step is silently exercising stale handler code** — including any future PR that changes `packages/hathor-rpc-handler/src/**` and expects the snap to pick it up.

**Suggested fix:** either add the build step to the "How to run" section, or (better) add a root `build`/`prepare` script (or a `pretest`/`pree2e` hook) so `yarn e2e` is self-sufficient from a clean checkout. A CI job — even a non-gating one — would have caught this immediately.

---

## 2. Determinism & the merge-bar contract for a non-gating suite

Empirically confirmed: even with the timeout hardening already in the suite, a single clean run on a slow headless Linux box was **2 passed / 3 failed** (~12 min). The non-determinism is real — the *same* testnet network-switch passed in `import` (39s) but hit the 180s test timeout as the first step of both `feature-example` and `token-lifecycle`.

[...]

PR #178, opened 2026-07-15
test(web-wallet): E2E follow-ups from #177 (PR 1 — light bundle)

raul-oliveira:
### Motivation

Follow-up improvements to the real-MetaMask E2E suite added in #169, tracked in #177. This is **PR 1 of 4** and lands the six low-risk items plus item 2's docs slice. **None of these change production runtime behavior.** The heavier items get their own PRs: de-flake (item 2b), nightly CI (item 2c), and Flask download checksum (item 6).

### Acceptance Criteria

Addresses these #177 items:

- **1** — `yarn e2e` is self-sufficient from a clean checkout: builds `@hathor/hathor-rpc-handler` first (only the handler — `snap-utils` is webpack-aliased to source).
- **3** — dApp page-object deadlines go through the centralized, `E2E_TIMEOUT_SCALE`-aware system (moved to `tests/e2e/timeouts.ts`, added `TIMEOUTS.dapp`; routed the 14 hardcoded `webWallet.ts` waits).
- **4** — the recovery-word read retries (bounded by `TIMEOUTS.seedReveal`) before failing, for the slow renders this suite targets.
- **5** — a unit test pins the `SNAP_TIMEOUT_MS` production default to 10s (empty→10s, valid override applies, invalid→10s).
- **7** — testnet-stub hygiene + dry-wallet top-up path documented.
- **8** — a loud guard fails on an unsupported MetaMask Flask version (`EXPECTED_FLASK_VERSION` + `assertFlaskVersion`, `E2E_ALLOW_FLASK_MISMATCH=1` escape hatch) plus an upgrade checklist.
- **2a** — the non-gating contract is documented where contributors look (`e2e.md`).
- Drive-by fix — `token-lifecycle.spec.ts` embedded `Date.now()`-derived token symbols in its test titles, so Playwright aborted the journey with "Test not found in the worker process"; titles are now static and the per-run symbols stay in the test bodies.

Out of scope (separate PRs): item 2b (de-flake), 2c (nightly CI), 6 (Flask download checksum).

### Validation

Run headed against real MetaMask Flask 13.31.0, inside `nix develop`:

- `yarn lint` — clean.
- `yarn test --run` — 268/268 (incl. the new `SNAP_TIMEOUT_MS` test; sanity-checked by breaking the default).
- `e2e:deps` rebuilds `dist` after removal (item 1).
- `yarn e2e` — all four journeys green: `onboarding` 9/9, `import` 2/2, `feature-example` 1/1, `token-lifecycle` 9 passed (+1 flaky-then-green). Cold-start connect flakes on some first attempts (item 2b's target) and clears on retry.
- item 8 mismatch → throws; escape hatch → warns (throwaway mocked test).

### Checklist

Refs #177


packages/web-wallet/tests/e2e/e2e.md
```diff
@@ -232,6 +238,16 @@ These bit us once each — check for them when a `getByRole`/`getByText` fails:
   `flex justify-between` row the label text and the toggle button are siblings, so from the
   label go up two levels (`xpath=ancestor::div[2]`) before selecting the button.
 
+## Two timeout knobs — which to reach for
+
+- **`E2E_TIMEOUT_SCALE`** (test side) multiplies every deadline in `tests/e2e/timeouts.ts` —
```
tuliomir, line 243: suggestion(non-blocking): file map still lists `timeouts.ts` under `driver/` after the move

Pinned to this changed line; the concern is the `## File map` block at `e2e.md:266`, outside the diff. It still nests `timeouts.ts` under `driver/`, and `tests/e2e/timeouts.ts` is missing from it — so the doc now contradicts the section right here. Since `e2e.md` opens with *Read this before reading or writing any E2E test*, that map is what the next contributor navigates by: they look under `driver/`, find nothing, and inline a magic number — the regression the move was meant to prevent.

Worth knowing a stale-path grep won't catch this; the map is an indented tree, so the path never appears as a literal string.

packages/web-wallet/tests/e2e/helpers/webWallet.ts
```diff
@@ -1,4 +1,5 @@
 import { expect, type Page } from '@playwright/test';
+import { TIMEOUTS } from '../timeouts';
```
tuliomir, line 2: suggestion(non-blocking): two connect-path deadlines escaped the centralization

Pinned to the new import; the sites are `waitForMetaMaskProvider(timeoutMs = 30_000)` at line 96 and the 8s `wallet_getSnaps` probe cap at line 83, both outside the diff. Neither scales with `E2E_TIMEOUT_SCALE`, and both sit on the cold-start connect path that item 2b targets — provider injection and the service-worker wake are exactly what run long on a slow machine. `waitForMetaMaskProvider` also swallows its own timeout (`.catch(() => undefined)`), so it degrades into a later, opaque failure rather than a clear one.

Suggest `TIMEOUTS.dapp.providerInject` / `.providerWarm`.

packages/web-wallet/tests/e2e/driver/MetaMaskDriver.ts
```diff
@@ -158,20 +161,38 @@ export class MetaMaskDriver {
   private async revealAndSaveSeed(): Promise<string[]> {
     const page = await this.openHome();
     await page.getByTestId(MetaMask.onboarding.recoveryReveal).click();
+
+    // Chips populate asynchronously; on a slow render a one-shot read can catch an empty chip
+    // and turn a transient into a hard failure. Poll the read (bounded by TIMEOUTS.seedReveal)
+    // until every word is present, THEN keep the fail-fast throw for a genuinely missing word.
+    const deadline = Date.now() + TIMEOUTS.seedReveal;
```
tuliomir, line 168: suggestion(non-blocking): `seedReveal` bounds the loop but not a single pass

The deadline is only checked *after* `readSeedWords` returns, and each chip read inherits `actionTimeout: 45_000` (`playwright.e2e.config.ts:67`). For attached-but-empty chips — the case your comment describes — this works as intended. If a chip isn't attached yet, one pass alone can burn 45s × 12 and blow the 180s test timeout, so Playwright kills the test generically and neither `dumpScreen('recovery-phrase-missing-word')` nor the actionable error message ever fires.

Passing a short per-read `{ timeout }` inside `readSeedWords` would make the 15s bound real.

packages/web-wallet/tests/e2e/driver/flask.ts
```diff
@@ -1,11 +1,45 @@
...
+ */
+function assertFlaskVersion(manifestDir: string): void {
+  let actual = '';
+  try {
+    const manifest = JSON.parse(readFileSync(join(manifestDir, 'manifest.json'), 'utf8')) as {
+      version?: string;
+    };
+    actual = manifest.version ?? '';
+  } catch {
+    return; // an unreadable/absent manifest is already handled by findManifestDir; don't double-fail
```
tuliomir, line 25: nitpick: an unparseable manifest silently bypasses the version guard

`findManifestDir` only tests `existsSync(manifest.json)` — it never parses it. So a truncated or half-extracted `.cache/flask-*/manifest.json` throws here and returns, skipping the guard entirely; that is one of the two scenarios the docstring above names (a stale `.cache`). Re-throwing, or at least warning, would keep the failure loud.

packages/web-wallet/tests/e2e/timeouts.ts
```diff
@@ -48,4 +53,29 @@ export const TIMEOUTS = {
     /** Hard cap for {@link MetaMaskDriver.driveApprovalsUntil} (runs until connected). */
     untilMaxMs: ms(170_000),
   },
+  /**
+   * dApp-side (page object) deadlines. These are the web-wallet's own UI waits — distinct from
+   * the driver's MetaMask deadlines above and from `SNAP_TIMEOUT_MS` (the dApp's 10s snap-RPC
+   * cap). Scaling `E2E_TIMEOUT_SCALE` now stretches these too.
+   */
+  dapp: {
```
tuliomir, line 61: nitpick: `driver/windows.ts` deadlines also sit outside the centralized system

Pinned here in the file they'd move into; the sites are `driver/windows.ts:25` (`timeout: 15_000`) and `:87` (`timeoutMs = 10_000`), outside the diff. Neither scales with `E2E_TIMEOUT_SCALE`. Lower stakes than the connect-path ones — this is the headed-only two-window split — but they're the last inline deadlines left in the E2E tree, so folding them in would let the docs claim `every deadline` without an asterisk.

packages/web-wallet/docs/qa-automation-strategy.md
```diff
@@ -163,6 +175,18 @@ swapping in your own):
...
+
+Selectors are pinned to one Flask build (`EXPECTED_FLASK_VERSION` in `driver/selectors.ts`); the
+resolver fails loudly on a mismatch (`E2E_ALLOW_FLASK_MISMATCH=1` overrides). To move the pin:
+
+1. Bump `E2E_FLASK_VERSION` (in `.env.e2e`) to the new release and clear `tests/e2e/.cache/`.
+2. Run with `E2E_ALLOW_FLASK_MISMATCH=1 yarn e2e:headed --project=onboarding` and walk the
+   onboarding + import screens, confirming each `data-testid` in `driver/selectors.ts`
+   (`MetaMask.onboarding.*`, `MetaMask.importing.*`, `Srp.*`, `Unlock.*`) still resolves.
+3. Update any changed selectors, then set `EXPECTED_FLASK_VERSION` to the new version.
+4. Update the pinned download checksum (see the Flask-integrity note, item 6 / PR 4).
```
tuliomir, line 188: nitpick: checklist step 4 points at a checksum that doesn't exist yet

Item 6 is explicitly out of scope for this PR and `driver/flask.ts` has no checksum today, so anyone following this checklist right now stalls on a step with nothing to update. Worth marking it as pending until that PR lands.

